PR #24, opened 2020-01-19
implement alternative fill rules for shapes

### LINKED ISSUE #11: Shapes: Please add support for Non-Zero rule so that i can do SVG rendering
imagesharp fills polygons using the even-odd rule where as the the default for svg is to use the nonzero-rule a fix would be required to change to Shapes to support the nonzero-rule plus a small change to Imagesharp to add to GraphicsOptions/ShapeRegion to allow passing thru the rule to use.

### LINKED ISSUE #14: Shapes: "ability to change fill type"
There was an open PR in Shapes: SixLabors/Shapes#68 by @equinox2k. If it's still something we need, than it shall be re-done it in the new repo, otherwise feel free to close this issue.

tocsoft:
### Prerequisites

### Description

Shapes/Drawing now supports both oddeven (old & default) and nonzero.

#### Nonzero 
Fill_IntersectionRules_Nonzero_Rgba32_Solid60x60_(0,0,255,255)

#### Oddeven (current)
Fill_IntersectionRules_OddEven_Rgba32_Solid60x60_(0,0,255,255)

Required me adding a new GraphicsOptions variant to expose the new shape specific properties. (inline with TextGraphicsOptions)  

resolves #14
resolves #11

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -20,7 +20,6 @@
   <!-- Default settings that explicitly differ from the Sdk.props defaults  -->
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <BaseIntermediateOutputPath>$(BaseArtifactsPath)obj/$(BaseArtifactsPathSuffix)/</BaseIntermediateOutputPath>
     <DebugType>portable</DebugType>
     <DebugType Condition="'$(codecov)' != ''">full</DebugType>
     <NullableContextOptions>disable</NullableContextOptions>
@@ -66,7 +65,6 @@
   <!-- Default settings that explicitly differ from the Sdk.targets defaults-->
   <PropertyGroup>
     <Authors>Six Labors and contributors</Authors>
-    <BaseOutputPath>$(BaseArtifactsPath)bin/$(BaseArtifactsPathSuffix)/</BaseOutputPath>
     <Company>Six Labors</Company>
     <PackageOutputPath>$(BaseArtifactsPath)pkg/$(BaseArtifactsPathSuffix)/$(Configuration)/</PackageOutputPath>
     <Product>SixLabors.ImageSharp.Drawing</Product>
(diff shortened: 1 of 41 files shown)


Directory.Build.props
```diff
@@ -20,7 +20,6 @@
   <!-- Default settings that explicitly differ from the Sdk.props defaults  -->
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <BaseIntermediateOutputPath>$(BaseArtifactsPath)obj/$(BaseArtifactsPathSuffix)/</BaseIntermediateOutputPath>
```
tocsoft: removed this because it was causing me issues while switching between VS and cmdline and it was cause things to get confused not even sure why you would want to move the obj/bin folders from the default locations

Directory.Build.props
```diff
@@ -66,7 +65,6 @@
   <!-- Default settings that explicitly differ from the Sdk.targets defaults-->
   <PropertyGroup>
     <Authors>Six Labors and contributors</Authors>
-    <BaseOutputPath>$(BaseArtifactsPath)bin/$(BaseArtifactsPathSuffix)/</BaseOutputPath>
```
tocsoft: as above

src/ImageSharp.Drawing/Processing/Processors/Drawing/FillRegionProcessor{TPixel}.cs
```diff
@@ -98,15 +99,13 @@ protected override void OnFrameApply(ImageFrame<TPixel> source)
...
                         {
-                            int pointsFound = region.Scan(subPixel + offset, buffer, configuration);
+                            int pointsFound = region.Scan(subPixel + offset, buffer, configuration, shapeOptions.IntersectionRule);
                             if (pointsFound == 0)
                             {
                                 // nothing on this line, skip
                                 continue;
                             }
 
-                            QuickSort.Sort(buffer.Slice(0, pointsFound));
```
tocsoft: we moved the sorting directly into to shape implementations thus no longer required here.

tocsoft:
@UkooLabs you will be happy to know I've finally managed to get to this one. Should solve your svg issues

UkooLabs:
> @UkooLabs you will be happy to know I've finally managed to get to this one. Should solve your svg issues

I am indeed very happy :)

Directory.Build.props
JimBobSquarePants: I've not seen any issues with command line building from root but am happy to go with this. I think it's a leftover from old dotnet build/test issues and the lack of solution level support.

PR #31, opened 2020-03-17
Update shared infrastructure

brianpopow:
### Prerequisites

### Description

Update shared infrastructure for the ImageSharp.Drawing project.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -91,8 +91,9 @@
       https://www.myget.org/F/sixlabors/api/v3/index.json;
       https://api.nuget.org/v3/index.json;
     </RestoreSources>
-    <SixLaborsPublicKey>002400000c8000009400000006020000002400005253413100040000010001000147e6fe6766715eec6cfed61f1e7dcdbf69748a3e355c67e9d8dfd953acab1d5e012ba34b23308166fdc61ee1d0390d5f36d814a6091dd4b5ed9eda5a26afced924c683b4bfb4b3d64b0586a57eff9f02b1f84e3cb0ddd518bd1697f2c84dcbb97eb8bb5c7801be12112ed0ec86db934b0e9a5171e6bb1384b6d2f7d54dfa97</SixLaborsPublicKey>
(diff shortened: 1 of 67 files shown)


brianpopow:
@JimBobSquarePants there are still some errors left in the test project like:

`##[error]Processing/FakeImageOperationsProvider.cs(12,50): error CS0281: Friend access was granted by 'SixLabors.ImageSharp, Version=1.0.0.0, Culture=neutral, PublicKeyToken=d998eea7b14cab13', but the public key of the output assembly ('') does not match that specified by the InternalsVisibleTo attribute in the granting assembly. `

from a quick google search, it seems related to strong naming ImageSharp. Unfortunately i have no experience with that. If you know more about that please let me know.

brianpopow:
Ok no more build errors, i have signed ImageSharp.Drawing.Tests.csproj with key from shared-infrastructure, see: 080acf2b0f4db675f0aa58e18dfacb5dddcdefa4. Still some failing tests. I will try to figure out whats wrong here.

JimBobSquarePants:
@brianpopow There's nothing in your changes that would cause differences. Could you post an example of two differing images?

@antonfirsov I'm a little concerned that https://github.com/SixLabors/ImageSharp/pull/1143 may have  caused more issues here.

brianpopow:
> @brianpopow There's nothing in your changes that would cause differences. Could you post an example of two differing images?

@JimBobSquarePants 
The failing test is in SolidFillBlendedShapesTests: `_1DarkBlueRect_2BlendHotPinkRect_3BlendTransparentEllipse`

This is the actual image output: 
actual

This is the expected one:
expected

They look the same but they are not, the square in the middle is different. Where the actual output has Rgba32(0, 0, 0, 0), the expected has Rgba32(255, 255, 255, 0).

Here is an image where the difference pixels are marked red:
output

JimBobSquarePants:
> Rgba32(0, 0, 0, 0), the expected has Rgba32(255, 255, 255, 0)

@brianpopow That makes me think that the expected result is actually incorrect. I wonder, do we premultiply in the pipeline anywhere?

brianpopow:
> > Rgba32(0, 0, 0, 0), the expected has Rgba32(255, 255, 255, 0)
> 
> @brianpopow That makes me think that the expected result is actually incorrect. I wonder, do we premultiply in the pipeline anywhere?

Premultiply is not used here. I dont think my changes have introduced this, so im keen in changing the reference images. @JimBobSquarePants should i do that?

JimBobSquarePants:
@brianpopow I figured it out! 

The tests use `Color.Transparent` which used to be `255,255,255,0` but is now `0,0,0,0`. 

Update the references :+1:

brianpopow:
> The tests use `Color.Transparent` which used to be `255,255,255,0` but is now `0,0,0,0`.

great, that explains its. 
Reference images are changed now.

JimBobSquarePants:
@brianpopow I fixed the build/signing issues for you also. Thanks for doing this, I'll merge it now.

brianpopow:
> @brianpopow I fixed the build/signing issues for you also. Thanks for doing this, I'll merge it now.

thanks, i could not figure out the problem

PR #36, opened 2020-04-11
Render glyphs with nonzero IntersectionRule

### LINKED ISSUE #35: Font renders with holes
Using ImageSharp 1.0.0-beta0002

I tried to render Windows 10 System font "Bahnschrift".

This is the result:

bahnschrift font

tocsoft:
### Prerequisites

### Description

Ensure we render glyphs using nonzero rule to prevent holes when using certain fonts.

Also fixed a side issue when starting trying to render a glyph completely to the left/above the image.

Fixes #35

### DIFF
--- a/src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
+++ b/src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
@@ -87,6 +87,17 @@ namespace SixLabors.ImageSharp.Drawing.Processing.Processors.Text
                             int startY = operation.Location.Y;
                             int startX = operation.Location.X;
                             int offsetSpan = 0;
+
+                            if (startX + buffer.Height < 0)
+                            {
+                                continue;
+                            }
+
+                            if (startX + buffer.Width < 0)
+                            {
+                                continue;
+                            }
+
                             if (startX < 0)
                             {
                                 offsetSpan = -startX;
@@ -335,7 +346,7 @@ namespace SixLabors.ImageSharp.Drawing.Processing.Processors.Text
                         {
                             var start = new PointF(path.Bounds.Left - 1, subPixel);
                             var end = new PointF(path.Bounds.Right + 1, subPixel);
-                            int pointsFound = path.FindIntersections(start, end, intersectionSpan);
+                            int pointsFound = path.FindIntersections(start, end, intersectionSpan, IntersectionRule.Nonzero);
 
                             if (pointsFound == 0)
                             {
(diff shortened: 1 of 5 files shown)


src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
```diff
@@ -87,6 +87,17 @@ void Draw(List<DrawingOperation> operations, IBrush brush)
                             int startY = operation.Location.Y;
                             int startX = operation.Location.X;
                             int offsetSpan = 0;
+
```
tocsoft, line 90: This fixes the issue where it fails when you try and render a glyph off image.

tests/ImageSharp.Drawing.Tests/Drawing/Text/DrawTextOnImageTests.cs
```diff
@@ -83,6 +83,22 @@ public void DoesntThrowExceptionWhenOverlappingRightEdge_Issue688_2<TPixel>(Test
         }
 
 
+        [Theory]
+        [WithSolidFilledImages(200, 200, "White", PixelTypes.Rgba32)]
+        public void OpenSansJWithNoneZeroShouldntExtendPastGlyphe<TPixel>(TestImageProvider<TPixel> provider)
```
tocsoft, line 88: This is tracking a regression the switch to nonzero caused

src/ImageSharp.Drawing/Shapes/ComplexPolygon.cs
```diff
@@ -190,21 +193,56 @@ public int FindIntersections(PointF start, PointF end, PointF[] buffer, int offs
...
         public int FindIntersections(PointF start, PointF end, Span<PointF> buffer, IntersectionRule intersectionRule)
         {
-            int totalAdded = 0;
-            for (int i = 0; i < this.paths.Length; i++)
+            if (this.internalPaths == null)
             {
-                Span<PointF> subBuffer = buffer.Slice(totalAdded);
-                int added = this.paths[i].FindIntersections(start, end, subBuffer, intersectionRule);
-                totalAdded += added;
+                lock (this.paths)
```
dlemstra, line 198: Why do you need the lock here?

tocsoft: because I expect this methods to be call in parallel and I only want the flattening and instantiation of the `InternalPath`s are relatively expensive operations.

I did however miss the 2nd null check to make sure I didn't actually run the code a 2nd time.

dlemstra: Maybe this should be moved to a separate method?

tocsoft: `FindIntersections()` is a method on the `IPath` interface and only `ComplexPolygon`s have this sort of initialization as its trying to efficiently composite the simplifed forms off all its child paths (which themselves can be `ComplexPolygons`)... not sure its worth adding all the complexity at this stage that change can probably be added when we review the API shapes.

JimBobSquarePants: I tend to agree. Get it working now so we have a reference to optimize and refactor against.

PR #45, opened 2020-05-13
Update licence to AGPL 3 & update Fonts dependency

tocsoft:
Mostly file header changes but also update to the public fonts dependency & drop the Compiler.Tooler dependency from the output package


tocsoft:
on appreciable change going on here so i'm just going to merge in anyway

PR #96, opened 2020-10-26
Fast polygon scanning with active edge list

### LINKED ISSUE #5: Fix rasterizing artifacts: offset vector operations inside `InternalPath` so they center around zero
To prevent issues around imprecise floating point maths on large numbers we should offset all points as they are processed into the `InternalPath` precalulation data so that all values are a close to zero as possible. Then after calling `InternalPath.FindIntersection` and similar methods we apply the reverse offset into the correct space the original points applied to.

The will reduce potential issues that can arise by the increased imprecision that occurs the further floating point values are away from zero.

SixLabors/Shapes#43 provided a bandaid fix by using double instead of a floats for some operations thus providing a simple fix for large shapes but offsetting will provide us with an algorithm that will support not just large shapes but ones with large offsets from zero.

### LINKED ISSUE #18: The performance of 'DrawLines' degrades the thinner the lines are
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp/issues) and [closed](https://github.com/SixLabors/ImageSharp/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Drawing thinner lines takes much more time to process than broader lines. 

The following tests were done on a 500x500 image (sorted by processing time):

- 100 lines with 10px: 448 ms
- 1000 lines with 10 px: 640 ms
- 100 lines with 1px: 1530 ms
- 1000 lines with 1px: 44680 ms

1000 10px lines render faster than 100 1px lines, which seems strange.

### Steps to Reproduce

**Option 1**

Create a simple project, referencing the latest ImageSharp library and include the code below:

```
var numberOfLines = 100;  // Update accordingly
var imageSize = 500; // Update accordingly
var lineWidth = 5; // Update accordingly

var lines = new List<int[]>();
var rnd = new Random(1);

for (int i = 0; i < numberOfLines; i++)
{
    int pixelX = rnd.Next(imageSize);
    int pixelY = rnd.Next(imageSize);
    lines.Add(new int[] { pixelX, pixelY });
}

SixLabors.Primitives.PointF[] imageSharpLines = lines
    .Select(r => new SixLabors.Primitives.PointF(r[0], r[1]))
    .ToArray();

PointF[] systemDrawingLines = lines.Select(r => new PointF(r[0], r[1])).ToArray();

var w = new Stopwatch();

w.Start();
using (var image = new Image<Rgba32>(imageSize, imageSize))
{
    image.Mutate(x => x

        .DrawLines(
            Rgba32.Red,
            lineWidth,
            imageSharpLines));

    using (var stream = File.Create($"test-imageSharp-{imageSize}-{numberOfLines}-{lineWidth}.png"))
    {
        image.SaveAsPng(stream);
    }
}
w.Stop();

Console.WriteLine($"Took {w.ElapsedMilliseconds} ms with ImageSharp");
```
**Option 2**

This was detected on a comparison between ImageSharp, System.Drawing and a custom simplified implementation. That benchmark is available at: https://github.com/pmcxs/core-linedrawing-benchmark/. It can be executed to replicate the issue:

- Clone the aforementioned repo
- Run it in Release. By default it will do various comparisons, including the ones specified above.

### System Configuration

- ImageSharp version: 1.0.0-dev002737
- Environment (Operating system, version and so on): Windows 10 Pro
- .NET Framework version: .NET Core 2.1

### LINKED ISSUE #61: Drawing text is too slow
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp.Drawing
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp.Drawing/issues) and [closed](https://github.com/SixLabors/ImageSharp.Drawing/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

(Wasn't sure if this should go here or in `SixLabors.Fonts`)

I'm trying to convert a working game engine from `system.drawing`, but my game engine (which usually hits 60 FPS without a problem) shows less than 10 FPS, due to a debug label which shows the cursor position and constantly updates as the cursor moves.

When I measured the text rendering time, I saw that `ImageSharp.Drawing` takes somewhat between 90ms and 170ms (median: 134ms) to render the text, and `system.drawing` takes ~0.3ms for rendering the text, so somewhat between x300 and x400 better performance.

Here's the analysis from xamarin profiler:
Screen Shot 2020-06-25 at 11 17 54 PM

`FindIntersectionWithOrientation` seems to be the biggest bottleneck: I'm completely ignorant when it comes to text rendering, I'm curious why we even need to find intersections when rendering text, I always imagined a glyph is just a series of path instructions that you map to size and render as is.

### Steps to Reproduce
- Run `DemoDesktop` from here: https://github.com/tzachshabtay/MonoAGS/tree/ImageSharp (using Visual Studio for Mac)
- Move your cursor around after the loading screen, and see the slowness
- The slow line is here: https://github.com/tzachshabtay/MonoAGS/blob/080bba5c186db4b88cb96e7c4940874cd5327102/Source/Engine/AGS.Engine/Misc/ImageSharp/ImageSharpTextDraw.cs#L89

If you want to compare to the `system.drawing` version, checkout the master branch, the equivalent line is here:
https://github.com/tzachshabtay/MonoAGS/blob/cadd7acbef6b1a97e47166681ba7e349074c4b47/Source/Engine/AGS.Engine.Desktop/Drawing/DesktopBitmapTextDraw.cs#L93

### System Configuration

- ImageSharp.Drawing version: 1.0.0-beta0009
- Other ImageSharp packages and versions: ImageSharp 1.0.0-rc0003
- Environment (Operating system, version and so on): Mac OSX Mojave 10.14
- .NET Framework version: `Mono JIT compiler version 6.4.0.208 (2019-06/07c23f2ca43 Wed Oct  2 04:52:23 EDT 2019)`

### LINKED ISSUE #90: Speed up scan-line rendering using "Active Edge Table"
Currently, brute force search is being used when calculating intersections, the scanline is being tested against all lines in the polygon for each row. This means `O(h*n)` intersection tests where `h` is the number of horizontal edges and `n` is the number of lines. 

The fastest way to achieve significant improvement with a (relatively) scoped change is to refactor the intersection detection logic to use an "Active Edge Table". In short this means:
- Sorting edges by both minimum `y` and maximum `y`
- Maintaining a list of edges which are currently intersected ("active") while iterating the scanlines
- .. so intersection tests are no longer needed, it's enough to calculate intersection points only for edges which are active.

This should reduce the number of intersection calculations to `O(n*log(n) + h*p)`, where `p` is the highest number of intersection points in a row during the scan.

Some sources describing the algorithm:
- https://www.techfak.uni-bielefeld.de/ags/wbski/lehre/digiSA/WS0607/3DVRCG/Vorlesung/13.RT3DCGVR-vertex-2-fragment.pdf
- https://www.geeksforgeeks.org/scan-line-polygon-filling-using-opengl-c/#:~:text=Scanline%20filling%20is%20basically%20filling,the%20vertices%20of%20the%20figure.

### Not in scope
Improving accuracy further (aka #5 and #15) is not my main goal here, the only criteria is to not regress. But I hope I can achieve better results, or at least get some insights.

### Plan
1. Expose an API on `IPath` returning tessellation results to be consumed by a horizontal scanline rasterizer. These should be stored in a memory-efficient, cache friendly form. A data structure equivalent to `(PointF, Orientation)[][]` could probably do the job, where `Orientation` is always assuming a horizontal scanline crossing the point.
2. Introduce a separate `ScanlineIterator` class that takes the tessellation, and exposes an API to iterate the horizontal scanlines.
3. Utilize `ScanlineIterator` in drawing code where we currently use `FindIntersections`, particularly: `FillRegionProcessor<T>`, `DrawTextProcessor<T>`

antonfirsov:
### Prerequisites

### Summary

This PR is intended to narrow the performance gap we have compared to other frameworks by implementing the algorithm described in #90.

~The PR is WIP, opened it to get early feedback. Remaining tasks:~
- [x] Review the output of 40+ failing tests with image comparer tool. Check if any of them indicates a bug VS minor difference in output.
- [x] Deal with any bugs found
- [x] Implement `NonZero` intersection rule in `PolygonScanner`
- [x] Utilize `PolygonScanner` in `DrawTextProcessor`
- [x] Dedupe `DrawTextProcessor` / `FillRegionProcessor` code
- [x] Deal with text drawing bugs
- [x] Remove the logic enforcing `TessellatedMultipolygon` orientation, and make sure the algorithm renders more or less nice output with CCW outermost contours + CW holes (see https://github.com/SixLabors/ImageSharp.Drawing/pull/96#issuecomment-723904465)
- [x] Remove sandbox code from both product and test (eg. `ShapeOptions.UsePolygonScanner`)
- [x] Cleanup and fix Style Cop findings

### Implementation

- The pipeline works as following: `IShape` -> `TesselatedMultipolygon` -> `ScanEdgeCollection` (compact representation of scannable edges) -> `PolygonScanner`
- `y` coordinates are rounded to the scanlines. This allows easy handling of corner cases. See [NumericCornerCases.jpg](https://raw.githubusercontent.com/SixLabors/ImageSharp.Drawing/3e3d18e827f5e3dfd60a8d74f2ae1a4750b34422/tests/ImageSharp.Drawing.Tests/Shapes/Scan/NumericCornerCases.jpg)
    - Horizontal lines are not emitted to `ScanEdgeCollection`. A line counts as horizontal by definition, if the whole line lies on the same scanline after rounding it's `y` coordinates.
- For more details about the algorithm, see comments, and the drawings pushed to the repo with documentation purposes.

### Results

Fixes #90 fixes #61 fixes #18.
Resolves #5 without offsetting the whole polygon, instead the calculation of coefficients in `ScanEdge` is centered around zero.

According to my visual observation, the new algorithm has fewer robustness issues now. I suspect the remaining ones are caused by an unhandled corner case rather than by numeric issue:

image

### Performance

The performance gains scale with the number of points in the polygon to draw. ( `~8.5x` speedup filling Utah's and `~62x` speedup filling Missisippi's polygons.)

```
Configuration:
BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-8650U CPU 1.90GHz (Kaby Lake R), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=5.0.100-rc.1.20452.10
  [Host]     : .NET Core 3.1.8 (CoreCLR 4.700.20.41105, CoreFX 4.700.20.41903), X64 RyuJIT
  DefaultJob : .NET Core 3.1.8 (CoreCLR 4.700.20.41105, CoreFX 4.700.20.41903), X64 RyuJIT


FillPolygonSmall (Utah): 

|                 Method |         Mean |      Error |     StdDev |  Ratio | RatioSD |
|----------------------- |-------------:|-----------:|-----------:|-------:|--------:|
|          SystemDrawing |    210.91 us |   4.130 us |   6.670 us |   4.72 |    0.19 |
|    ImageSharp_Original | 28,572.78 us | 350.107 us | 292.356 us | 629.88 |   18.67 |
| ImageSharp_ActiveEdges |  3,342.78 us |  63.592 us |  68.043 us |  74.06 |    2.59 |
|              SkiaSharp |     44.79 us |   0.825 us |   1.184 us |   1.00 |    0.00 |


FillPolygonMedium (Mississippi):

|                 Method |         Mean |       Error |      StdDev |    Ratio | RatioSD |
|----------------------- |-------------:|------------:|------------:|---------:|--------:|
|          SystemDrawing |     358.2 us |     7.14 us |    10.68 us |     2.27 |    0.06 |
|    ImageSharp_Original | 187,078.3 us | 3,699.54 us | 6,181.11 us | 1,186.12 |   37.40 |
| ImageSharp_ActiveEdges |   2,884.7 us |    56.27 us |   109.75 us |    18.84 |    0.67 |
|              SkiaSharp |     156.8 us |     2.91 us |     2.43 us |     1.00 |    0.00 |


DrawPolygonMediumThin:


|                 Method |         Mean |       Error |       StdDev |    Ratio | RatioSD |
|----------------------- |-------------:|------------:|-------------:|---------:|--------:|
|          SystemDrawing |     697.7 us |    13.94 us |     25.49 us |     6.08 |    0.27 |
|    ImageSharp_Original | 407,472.0 us | 7,985.96 us | 10,931.27 us | 3,541.83 |  117.29 |
| ImageSharp_ActiveEdges |  18,913.2 us |   327.80 us |    350.75 us |   165.31 |    3.56 |
|              SkiaSharp |     114.6 us |     2.11 us |      1.98 us |     1.00 |    0.00 |


DrawPolygonMediumThick:


|                 Method |       Mean |     Error |    StdDev |     Median |  Ratio | RatioSD |
|----------------------- |-----------:|----------:|----------:|-----------:|-------:|--------:|
|          SystemDrawing |   1.539 ms | 0.0304 ms | 0.0621 ms |   1.515 ms |   0.97 |    0.04 |
|    ImageSharp_Original | 169.697 ms | 3.3095 ms | 4.7463 ms | 168.241 ms | 104.43 |    2.09 |
| ImageSharp_ActiveEdges |  38.795 ms | 0.7716 ms | 1.0562 ms |  38.591 ms |  24.16 |    0.74 |
|              SkiaSharp |   1.625 ms | 0.0158 ms | 0.0132 ms |   1.623 ms |   1.00 |    0.00 |

```


src/ImageSharp.Drawing/Primitives/Region.cs
```diff
@@ -23,6 +23,8 @@ public abstract class Region
         /// </remarks>
         public abstract Rectangle Bounds { get; }
 
+        public abstract IPath Shape { get; }
```
antonfirsov, line 26: @tocsoft `Region` seems to be a thin wrapper around `IPath` in practice. Do we really need it?

antonfirsov:
The `FillPolygonSmall` benchmark was broken. After the fix, there is  `~8.5x` speedup filling Utah's and `~62x` speedup filling Missisippi's polygons.

src/ImageSharp.Drawing/Primitives/Region.cs
tocsoft: My logic for the `Region` abstraction was so that it can live in the main ImageSharp repo so that we can eventually be able to support clipping across the library... at the moment we have a percentage of our APIs taking a rectangle to apply an effect too, I instead wanted to be able to pass a `Region` to apply the clipping effect to for better/simpler/parity with other graphics libraries.
 support.

antonfirsov: Makes sense, but if I understand your approach correctly, it's quite an expensive goal:
1. `Region` shall be moved to the main library + a new efficient abstraction defined for returning scannable data. <- this work shall be milestoned to 1.0 of Drawing.
2. Processors shall be refactored to have a code path that consumes `Region` and does scanning with multisampling. For performance, this should be probably a separate code path than the standard `Rectangle` path, resulting in duplication. (-> maintenance pain.)

There is a simpler, and probably more CPU-friendly (although less memory-friendly) way to achieve the same:
A. Render the effect onto a separate image and `FillRegion` the overlay image onto the main one.
B. Have a universal helper for A. + extension method overloads for each applicable main library processor, taking `Region` and/or `IPath` instead of `Rectangle`. These extension methods can live in both ImageSharp and ImageSharp.Drawing (depending on whether we want to invest into 1.)

src/ImageSharp.Drawing/Shapes/Scan/ClassicPolygonScanner.cs
```diff
@@ -0,0 +1,92 @@
+using System;
+using System.Buffers;
+using System.Runtime.CompilerServices;
+using SixLabors.ImageSharp.Memory;
+
+namespace SixLabors.ImageSharp.Drawing.Shapes.Scan
+{
+    internal ref struct ClassicPolygonScanner
```
antonfirsov, line 8: This is just a temporary thing. Helped me to refactor stuff in smaller steps.

src/ImageSharp.Drawing/Primitives/Region.cs
tocsoft: hmm... now you say that that is basically what I did in this sample https://github.com/SixLabors/Samples/blob/master/ImageSharp/CustomImageProcessor/Program.cs maybe we don't need to bother and instead add a extension to ImageSharp.Drawing i.e. `imageContext.Clip(i=>i.DrawText(""))`; makes me think we should just drop the type entirely after all if clipping is a IS.Drawing extension.

antonfirsov: Yeah, I believe this approach will simplify a lot, and would achieve the same with significantly much less amount of work.

antonfirsov:
I realized that my CAD experience fooled me, since computational geometry algorithms usually depend on specific orientation being enforced for polygons.  I made a big mistake by not reading up the exact definition of even/odd and nonzero rules making the scanning algorithm depend on shape orientation, which is [enforced](https://github.com/SixLabors/ImageSharp.Drawing/pull/96/files#diff-a6cd16d5819be1b13d7121bb8a043b61861b2c3f9f1cab944e1630f839fa4e9eR130) while creating `TessellatedMultipolygon`.

I used the orientation to define more accurate rules for corner cases where a horizontal contour line is being intersected. Vast majority of real-life complex polygons & multi-polygons will be defined with outermost contours being clockwise and holes being CCW (screen space), so it's enough to omit the logic enforcing pre-defined orientations. But now I also need to apply modifications to my algorithm, so it works properly in cases where input polygons are oriented the opposite way.

antonfirsov:
Ready for review. The algorithm is explained in:
https://github.com/SixLabors/ImageSharp.Drawing/blob/af/polygon-lab/src/ImageSharp.Drawing/Shapes/Rasterization/PolygonScanning.MD

tests/ImageSharp.Drawing.Benchmarks/Drawing/DrawTextOutline.cs
```diff
@@ -115,5 +115,23 @@ public void DrawTextCoreOld()
...
+        // |-------------- |--------------- |-------------:|-----------:|-----------:|-------:|--------:|----------:|---------:|------:|----------:|
+        // | SystemDrawing |              1 |     55.03 us |   0.199 us |   0.186 us |   5.43 |    0.03 |         - |        - |     - |      40 B |
+        // |    ImageSharp |              1 |  2,161.92 us |   4.203 us |   3.510 us | 213.14 |    0.52 |  253.9063 |        - |     - |  804452 B |
+        // |     SkiaSharp |              1 |     10.14 us |   0.040 us |   0.031 us |   1.00 |    0.00 |    0.5341 |        - |     - |    1680 B |
+        // |               |                |              |            |            |        |         |           |          |       |           |
+        // | SystemDrawing |             20 |  1,450.12 us |   3.583 us |   3.176 us |  27.36 |    0.11 |         - |        - |     - |    3696 B |
+        // |    ImageSharp |             20 | 28,559.17 us | 244.615 us | 216.844 us | 538.85 |    3.98 | 2312.5000 | 781.2500 |     - | 9509056 B |
+        // |     SkiaSharp |             20 |     53.00 us |   0.166 us |   0.147 us |   1.00 |    0.00 |    1.6479 |        - |     - |    5336 B |
```
antonfirsov, line 135: Unless I'm doing something wrong in this benchmark. We are still a lot behind when it comes to draw text.

JimBobSquarePants: Hmmm... Anything obvious when profiling?

src/ImageSharp.Drawing/Shapes/Rasterization/PolygonScanner.cs
```diff
@@ -0,0 +1,235 @@
...
+            OrientationHandling orientationHandling = OrientationHandling.FirstRingIsContourFollowedByHoles)
+        {
+            TessellatedMultipolygon multipolygon = TessellatedMultipolygon.Create(polygon, allocator, orientationHandling);
+            ScanEdgeCollection edges = ScanEdgeCollection.Create(multipolygon, allocator, subsampling);
+            PolygonScanner scanner = new PolygonScanner(edges, multipolygon.TotalVertexCount * 2, minY, maxY, subsampling, intersectionRule, allocator);
+            scanner.Init();
+            return scanner;
+        }
+
+        public static PolygonScanner Create(
```
tocsoft, line 109: this is only called from one place, we should probably just call the above `Create()` method directly

src/ImageSharp.Drawing/Shapes/Rasterization/PolygonScanner.cs
```diff
@@ -0,0 +1,235 @@
...
+            IPath polygon,
+            int minY,
+            int maxY,
+            int subsampling,
+            IntersectionRule intersectionRule,
+            MemoryAllocator allocator,
+            OrientationHandling orientationHandling = OrientationHandling.FirstRingIsContourFollowedByHoles)
```
tocsoft, line 100: all call sites set this so don't really want a default set.

All call sites (except tests) seem to always end up passing the the same option `ForcePositiveOrientationOnSimplePolygons` do we need the other options? and thus are the tests actually needed/wanted if no production code uses the setting?

src/ImageSharp.Drawing/Shapes/TessellatedMultipolygon.cs
```diff
@@ -0,0 +1,161 @@
...
+        public int TotalVertexCount { get; }
+
+        public int Count => this.rings.Length;
+
+        public Ring this[int index] => this.rings[index];
+
+        public static TessellatedMultipolygon Create(
+            IPath path,
+            MemoryAllocator memoryAllocator,
+            OrientationHandling orientationHandling = OrientationHandling.ForcePositiveOrientationOnSimplePolygons)
```
tocsoft, line 46: `ForcePositiveOrientationOnSimplePolygons` ends up not actually influencing the algorithm do we need the extra complexity of the other types seeing as none of our library code uses anything but this option

tests/ImageSharp.Drawing.Benchmarks/Drawing/DrawTextOutline.cs
antonfirsov: Code is same as with `FillRegion`. I suppose GDI and Skia are utilizing some further optimizations. Probably a global glyph cache?

tocsoft (commented):
only thing I can see is the fact we don't seem have any need at all for any for the `OrientationHandling` logic which leaves us with additional maintenance.

src/ImageSharp.Drawing/Shapes/TessellatedMultipolygon.cs
antonfirsov: I just wanted to keep the flexibility to change this later if we want. Also: having the 3 options kinda documents what are we doing here and why.

src/ImageSharp.Drawing/Utilities/NumericUtilities.cs
```diff
@@ -0,0 +1,102 @@
...
+        // since we do target only 3.1+, we can detect it's presence by using SUPPORTS_RUNTIME_INTRINSICS
+        // TODO: Ideally this should have a separate definition in Build.props, but that adaption should be done cross-repo. Using a workaround until then.
+#if SUPPORTS_RUNTIME_INTRINSICS
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int Log2(uint value) => System.Numerics.BitOperations.Log2(value);
+
+#else
+
+#pragma warning disable SA1515, SA1414, SA1114, SA1201
+        private static ReadOnlySpan<byte> Log2DeBruijn => new byte[32]
```
JimBobSquarePants, line 53: Some cool stuff in this class 👍 I swear I saw @saucecontrol optimize this in the runtime for NET5

src/ImageSharp.Drawing/Utilities/NumericUtilities.cs
```diff
@@ -0,0 +1,102 @@
...
+                // Using deBruijn sequence, k=2, n=5 (2^5=32) : 0b_0000_0111_1100_0100_1010_1100_1101_1101u
+                ref MemoryMarshal.GetReference(Log2DeBruijn),
+                // uint|long -> IntPtr cast on 32-bit platforms does expensive overflow checks not needed here
+                (IntPtr)(int)((value * 0x07C4ACDDu) >> 27));
+        }
+
+#pragma warning restore
+#endif
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static float ClampFloat(float value, float min, float max)
```
JimBobSquarePants, line 87: I have a dream of one day converting all the extension methods that do this in the core repo to a utils class. I hate that they are extensions.

src/ImageSharp.Drawing/Utilities/SortUtility.KeyValueSort.cs
```diff
@@ -0,0 +1,204 @@
...
+using System.Numerics;
+using System.Runtime.CompilerServices;
+
+namespace SixLabors.ImageSharp.Drawing.Utilities
+{
+    internal static partial class SortUtility
+    {
+        // Adapted from:
+        // https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/ArraySortHelper.cs
+        // If targeting .NET 5, we can call span based sort, but probably not worth it only for that API.
```
JimBobSquarePants, line 14: I imagine we'll target NET5 at some point. I plan to for the main repo for the AMD intrinsics.

JimBobSquarePants (commented):
It all looks good to me, just a few questions really. I'll leave @tocsoft to do final approval as he actually understands this stuff! 😝

src/ImageSharp.Drawing/Shapes/InternalPath.cs
```diff
@@ -333,25 +337,25 @@ public int FindIntersectionsWithOrientation(PointF start, PointF end, Span<Point
                         Vector2 point = FindIntersection(this.points[i].Segment, target);
```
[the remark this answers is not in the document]
antonfirsov: This method is no longer used for intersection calculation in rasterizer (replaced by `PolygonScanner` and `ScanEdge`).

I'm proposing to remove `IPath.FindIntersections()` as a feature in #103

src/ImageSharp.Drawing/Utilities/SortUtility.KeyValueSort.cs
antonfirsov: We need an issue/discussion for that then, there are many aspects to investigate.

src/ImageSharp.Drawing/Shapes/InternalPath.cs
```diff
@@ -459,7 +463,7 @@ public bool PointInPolygon(PointF point)
                 // check if the point is on an intersection is it is then inside
                 for (int i = 0; i < intersection; i++)
                 {
-                    if (buffer[i].Equivelent(point, Epsilon))
+                    if (buffer[i].Equivalent(point, Epsilon))
```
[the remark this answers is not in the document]
antonfirsov: Based on the code around, no. This change only a rename because of a typo, this `PointInPolygon` check is unused by the code in the PR.

src/ImageSharp.Drawing/Shapes/Rasterization/ActiveEdgeList.cs
```diff
@@ -0,0 +1,289 @@
...
+        {
+            this.count = 0;
+            this.Buffer = buffer;
+        }
+
+        private Span<int> ActiveEdges => this.Buffer.Slice(0, this.count);
+
+        public void EnterEdge(int edgeIdx)
+        {
+            this.Buffer[this.count++] = edgeIdx | EnteringEdgeFlag;
```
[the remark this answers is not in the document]
antonfirsov: Yes. Although not worth of further microoptimization IMO. Focus should be moved to #102

src/ImageSharp.Drawing/Shapes/Rasterization/ScanEdgeCollection.Build.cs
```diff
@@ -0,0 +1,310 @@
...
+            }
+
+            static void RoundY(ReadOnlySpan<PointF> vertices, Span<float> destination, float subsamplingRatio)
+            {
+                for (int i = 0; i < vertices.Length; i++)
+                {
+                    // for future SIMD impl:
+                    // https://www.ocf.berkeley.edu/~horie/rounding.html
+                    // Avx.RoundToPositiveInfinity()
+                    destination[i] = MathF.Round(vertices[i].Y * subsamplingRatio, MidpointRounding.AwayFromZero) / subsamplingRatio;
```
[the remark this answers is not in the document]
antonfirsov: Not significant ATM, will be only bottleneck after #102 is addressed.

src/ImageSharp.Drawing/Utilities/NumericUtilities.cs
antonfirsov: I have no idea what does this code do 😅 Copypasted it so the SortUtility stuff compiles agains older frameworks.

src/ImageSharp.Drawing/Processing/Processors/Drawing/FillRegionProcessor{TPixel}.cs
```diff
@@ -54,127 +55,86 @@ protected override void OnFrameApply(ImageFrame<TPixel> source)
...
-                                    {
-                                        scanline[x] = 1;
-                                        hasOnes = true;
-                                    }
-                                    else
-                                    {
-                                        scanline[x] = 0;
-                                        hasZeros = true;
-                                    }
+                                    source.GetPixelRowSpan(y).Slice(minX, scanlineWidth).Fill(solidBrushColor);
```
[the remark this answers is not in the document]
antonfirsov: Each call `scanner.MoveToNextPixelLine()` is moving to the next `y`:

https://github.com/SixLabors/ImageSharp.Drawing/blob/6d1e7a10d751ecce68b6e9845cbf54bcdcf7f6cf/src/ImageSharp.Drawing/Processing/Processors/Drawing/FillRegionProcessor%7BTPixel%7D.cs#L88

There are no other loops around this call. I don't know when is `isSolidBrushWithoutBlending == true`  in practice.

src/ImageSharp.Drawing/Utilities/NumericUtilities.cs
antonfirsov: Even better if they are available as bulk operations.

src/ImageSharp.Drawing/Shapes/InternalPath.cs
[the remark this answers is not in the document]
JimBobSquarePants: Hurrah!

src/ImageSharp.Drawing/Shapes/Rasterization/ActiveEdgeList.cs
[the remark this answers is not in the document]
JimBobSquarePants: Yeah, I had a look at the blender code after you raised the issue. Ripe for optimization!

src/ImageSharp.Drawing/Utilities/NumericUtilities.cs
JimBobSquarePants: Super agree.

JimBobSquarePants (approved):
Very happy!

tocsoft (dismissed):
Agreed everything's looking good.... still not 100% convinced we want the extra (unused) logic in `TessellatedMultipolygon` but I'm happy to approve anyway. But I agree there is some follow up clean up actions to remove some unused/wanted codepaths, pretty much all of internal path can probably be dropped along with the more generic point in polygon/intersection logic etc from the paths.

src/ImageSharp.Drawing/Utilities/SortUtility.KeyValueSort.cs
JimBobSquarePants: Have added https://github.com/SixLabors/ImageSharp/issues/1432

antonfirsov:
@tocsoft I will remove the `OrientationHandling` stuff as part of this PR in the next few days.

antonfirsov:
@tocsoft all done, I'm letting you have a final look and/or merge.

tocsoft (approved):
approved 👍  nice work

PR #99, opened 2020-11-02
Improve performance of path gradient brush on triangles.

maxvoxel8:
### Prerequisites

- [X] I have written a descriptive pull-request title
- [X] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp.Drawing/pulls) open
- [X] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [X] I have provided test coverage for my change (where applicable)

### Description

I have found that the performance of PathGradientBrush is not very good. Since in my application I am only filling triangles, I thought I'd fix it for this specific case. This reduces the time spent in the pixel getter from 50.2s to 4.7s (10.7x improvement).

In the general case, the right way would probably be to triangulate the polygon and then apply the triangle method to the individual triangles. However there is no convenient triangulation library available right now (though the clipper author is working on one).


src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
```diff
@@ -257,6 +257,25 @@ private class PathGradientBrushApplicator<TPixel> : BrushApplicator<TPixel>
...
+                                                      out float u,
+                                                      out float v))
+                        {
+                            return this.transparentPixel;
+                        }
+
+                        Vector4 pointColor = ((1 - u - v) * this.edges[0].StartColor) + (u * this.edges[0].EndColor) +
+                                         (v * this.edges[2].StartColor);
+
+                        return new Color(pointColor).ToPixel<TPixel>();
```
antonfirsov, line 276:
```suggestion
                        TPixel pixel = default;
                        pixel.FromVector4(pointColor);
                        return pixel;
```

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
```diff
@@ -301,6 +320,40 @@ private class PathGradientBrushApplicator<TPixel> : BrushApplicator<TPixel>
...
+
+                Vector2 pv1 = point - v1;
+                Vector2 pv2 = point - v2;
+                Vector2 pv3 = point - v3;
+
+                var d1 = Vector3.Cross(new Vector3(e1.X, e1.Y, 0), new Vector3(pv1.X, pv1.Y, 0));
+                var d2 = Vector3.Cross(new Vector3(e2.X, e2.Y, 0), new Vector3(pv2.X, pv2.Y, 0));
+                var d3 = Vector3.Cross(new Vector3(e3.X, e3.Y, 0), new Vector3(pv3.X, pv3.Y, 0));
+
+                if (Math.Sign(Vector3.Dot(d1, d2)) != Math.Sign(Vector3.Dot(d1, d3)) || Math.Sign(Vector3.Dot(d1, d2)) != Math.Sign(Vector3.Dot(d2, d3)))
```
antonfirsov, line 338: This could be a source of inaccuracies. Near-zero should probably count as zero, I would experiment with tolerant comparison.

antonfirsov (commented):
Thanks! I think this is a nice addition.

Looking at the new reference outputs, I think the old ones looked better. How does the before/after difference look like for bigger triangles (eg. 20x20, 50x50)? Any chance you can improve the accuracy?

I think you can achieve the same optimization for an arbitrary polygon by applying a simple  winding number point in polygon test. Maybe you want to give it a try?

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
maxvoxel8: This now matches the original output so I think it's good.

maxvoxel8:
> Looking at the new reference outputs, I think the old ones looked better. How does the before/after difference look like for bigger triangles (eg. 20x20, 50x50)? Any chance you can improve the accuracy?

I increased the test to 20x20 and changed the edge test, there is now no difference relative to master.

> I think you can achieve the same optimization for an arbitrary polygon by applying a simple winding number point-in-polygon test. Maybe you want to give it a try?

The optimization is not only about checking if the point is in the polygon, it's faster in all cases.

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
```diff
@@ -257,6 +264,27 @@ private class PathGradientBrushApplicator<TPixel> : BrushApplicator<TPixel>
...
+                            return this.transparentPixel;
+                        }
+
+                        Vector4 pointColor = ((1 - u - v) * this.edges[0].StartColor) + (u * this.edges[0].EndColor) +
+                                         (v * this.edges[2].StartColor);
+
+                        TPixel px = default;
+                        px.FromVector4(pointColor);
+                        return px;
+                    }
```
antonfirsov, line 287: Can you try if the new tests are also passing after commenting out this change? This would proove that the inputs are identical on bigger scale. In that case I'm fine with the change, but @tocsoft should approve too.

maxvoxel8:
Actually it looks like your suggested pixel conversion doesn't always work. I'm using the HalfSingle pixel type. If the `pointColor` vector is `<0, 0, 0, 1>`, `new Color(pointColor).ToPixel<TPixel>()` correctly returns `HalfSingle(-1)`, while

```
TPixel px = default;
px.FromVector4(pointColor);
```
returns `HalfSingle(0)`.

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
maxvoxel8: The tests all pass, just code coverage is failing: https://github.com/SixLabors/ImageSharp.Drawing/pull/99/commits/b7b5d7296a67944b62de08cd7ef946df44d42fc5

antonfirsov: Cool, thanks, can revert now!

antonfirsov:
@maxvoxel8 https://github.com/SixLabors/ImageSharp.Drawing/pull/99#issuecomment-721276955 should be a bug in `HalfSingle.FromVector4()` or `HalfSingle.FromRgba64()` depending on whether `HalfSingle(-1)` or `HalfSingle(0)` is the correct value (I've no idea). Would be very helpful if you could file an issue in the main ImageSharp repo:
https://github.com/SixLabors/ImageSharp

In the meanwhile, feel free to revert to your original conversion, adding a comment about the bug to make sure no one will "optimize" away your workaround before the `HalfSingle` issue is fixed.

antonfirsov:
> In the meanwhile, feel free to revert to your original conversion

Ah I see, it's already done. Comment + ImageSharp issue would be still nice.

maxvoxel8:
Done. The test that I added, FillTriangleWithGreyscale, will also catch if the color conversion is incorrect.

JimBobSquarePants:
I think we're mixing up `ScaledVector4` and `Vector4` here.

antonfirsov:
@JimBobSquarePants ah right, might be the case.

@maxvoxel8 can you try if my suggestion works with `FromScaledVector4` ? In that case SixLabors/ImageSharp#1412 is invalid and we can close it.

maxvoxel8:
Yes, that works. One gradient test had a 0.0137% error, I updated the image.

antonfirsov (approved):
Although the optimization addresses a very specific case, it does not harm the generic path, and it gives ideas how to improve further, so I'm fine with this change.

Let's see what others say.

JimBobSquarePants:
@antonfirsov if you're happy, I'm happy. 👍

antonfirsov:
Then let's merge, @maxvoxel8 thanks again for taking the time for this!

PR #111, opened 2020-12-03
Use Shared props/targets

JimBobSquarePants:
### Prerequisites

### Description

- Updates the solution to use the shared props/targets
- Cleans up all warnings in the solution.
- Adds Mac OSX CI workflow plus additional enhancements from ImageSharp.


antonfirsov (changes requested):
Didn't get to drawing yet, but I see 999 build warnings in main lib's ImageSharp.Test project per build target. We should fix that before propagating the rules further.

Most of them are "this should not follow that" and underscore rules we do not wan to apply in test code.

JimBobSquarePants:
>Didn't get to drawing yet, but I see 999 build warnings in main lib's ImageSharp.Test project per build target. We should fix that before propagating the changes further.

I don't think your submodule is loaded properly or something. I brought it down to zero from over 1200 warnings.

antonfirsov:
@JimBobSquarePants ah sorry you may be right will doublecheck tomorrow!

antonfirsov (approved):
@JimBobSquarePants this ImageSharp.Drawing PR is fine, but still I get those warnings with the main library's test project even on a fresh clone on master. (After standard `git submodule update --init --recursive`).

Nice job cleaning up the Drawing project, and the overall effort across the SixLabors infra. No idea why is it OK here and not OK there.

antonfirsov:
Opened SixLabors/ImageSharp#1460 to properly track the problem.

PR #126, opened 2021-03-24
Apply global matrix transform drawing text and shapes.

### LINKED ISSUE #125: Allow passing Matrix3x2 to DrawTextProcessor
Follow-up on the discussion #124.

We should enable using an arbitrary `Matrix3x2` transform as a parameter within [`DrawTextProcessor`](https://github.com/SixLabors/ImageSharp.Drawing/blob/799d6afce0749f3316f6e766ac8884cccc9ef46b/src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor.cs).

The easiest implementation seems to be to pass it to `PathBuilder` here:
https://github.com/SixLabors/ImageSharp.Drawing/blob/e09fe9b91877cf0ae61480dee087476d0208d29d/src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor%7BTPixel%7D.cs#L223

### Open questions
Before implementing, we need to figure out how does the transform interfere with [`DrawTextProcessor.Location`](https://github.com/SixLabors/ImageSharp.Drawing/blob/799d6afce0749f3316f6e766ac8884cccc9ef46b/src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor.cs#L71) and ultimately [`RenderOptions.Origin`](https://github.com/SixLabors/Fonts/blob/5f9cea1ba0d6c42488fdfca946cbaebce788e2bf/src/SixLabors.Fonts/RendererOptions.cs#L147). Ideally, I would replace `DrawTextProcessor.Location` with `DrawTextProcessor.Transform` entirely, but there might be things I'm missing.

@tocsoft any thoughts?

tocsoft:
### Prerequisites

### Description

Adds a `Transform` property to both `ShapeOptions` and `TextOptions` (which can be registered globally).

This allows for usages using this syntax (for shapes):
```
var matrix = Matrix3x2.CreateSkew(GeometryUtilities.DegreeToRadian(-15), 0, new Vector2(200, 200));
image.Mutate(c=> c.SetShapeOptions(x => x.Transform = matrix)
                .FillPolygon(Color.White, simplePath));
```

or for text

```
var matrix = Matrix3x2.CreateSkew(GeometryUtilities.DegreeToRadian(-15), 0, new Vector2(200, 200));
image.Mutate(c=> c.SetTextOptions(x => x.Transform = matrix)
                .DrawText(text, font, Color.White, new PointF(10, 10)));
```

Current limitations on this pattern is that it uses 2 separate properties... Ideally we would add the `Transform` property to `GraphicsOptions` (which is available for both shapes and text processors) but that class lives in the main ImageSharp repo so would require a change over there.. if we are happy to make the change over there then we should do that first then update this PR to set it centrally. 

resolves #125


src/ImageSharp.Drawing/Processing/Processors/Drawing/DrawPathProcessor.cs
```diff
@@ -45,7 +45,7 @@ public IImageProcessor<TPixel> CreatePixelSpecificProcessor<TPixel>(Configuratio
             where TPixel : unmanaged, IPixel<TPixel>
         {
             // offset drawlines to align drawing outlines to pixel centers
-            var shape = this.Shape.Translate(0.5f, 0.5f);
+            var shape = this.Shape.Translate(0.5f, 0.5f).Transform(this.Options.ShapeOptions.Transform);
```
antonfirsov, line 48: [Perf] We should avoid doing a full Transform / Copy of the polygon in the default case:

```suggestion
            var shape = this.Shape.Translate(0.5f, 0.5f);
            if (this.Options.ShapeOptions.Transform != Matrix3x2.Identity)
            {
                shape = shape.Transform(this.Options.ShapeOptions.Transform);
            }
```

src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
```diff
@@ -367,7 +377,7 @@ public void EndGlyph()
 
             private Buffer2D<float> Render(IPath path)
             {
-                Size size = Rectangle.Ceiling(path.Bounds).Size;
+                Size size = Rectangle.Ceiling(path.Bounds).Size * 2;
```
antonfirsov, line 380: Why `* 2` here?

src/ImageSharp.Drawing/Processing/Processors/Drawing/FillPathProcessor.cs
```diff
@@ -45,7 +45,9 @@ public FillPathProcessor(ShapeGraphicsOptions options, IBrush brush, IPath shape
         public IImageProcessor<TPixel> CreatePixelSpecificProcessor<TPixel>(Configuration configuration, Image<TPixel> source, Rectangle sourceRectangle)
             where TPixel : unmanaged, IPixel<TPixel>
         {
-            if (this.Shape is RectangularPolygon rectPoly)
+            var shape = this.Shape.Transform(this.Options.ShapeOptions.Transform);
```
antonfirsov, line 48: Same as above.

tests/ImageSharp.Drawing.Tests/Drawing/Text/DrawTextOnImageTests.cs
```diff
@@ -158,6 +159,81 @@ public void OpenSansJWithNoneZeroShouldntExtendPastGlyphe<TPixel>(TestImageProvi
...
+            provider.RunValidatingProcessorTest(
+                c => c
+                    .SetTextOptions(o =>
+                    {
+                        o.HorizontalAlignment = HorizontalAlignment.Center;
+                        o.VerticalAlignment = VerticalAlignment.Center;
+                        o.Transform = Matrix3x2.CreateRotation(radians, new Vector2(rotationOriginX, rotationOriginY));
+                    })
+                    .DrawText(text, font, Color.Black, new PointF(x, y)),
+                $"{fontName}-{fontSize}-{ToTestOutputDisplayText(text)}-({x},{y})",
```
antonfirsov, line 192: `angle` is missing, adding a test case with stuff other than 45° will mess up the test:

```suggestion
                $"F({fontName})-S({fontSize})-A({angle})-{ToTestOutputDisplayText(text)}-({x},{y})",
```

JimBobSquarePants:
@tocsoft Are there any issues you see with adding the property upstream? I.E do you think people will try to use it for non related operations?

antonfirsov (commented):
I don't think we should expose transform in the core library's `GraphicsOptions`, will elaborate in a follow-up comment.

As for this PR, I think we should merge it after fixing a few minor things, and iterate on the API later.

antonfirsov:
I would love to see the duplication between `TextGraphicsOptions` and `ShapeGraphicsOptions` being removed, but I don't think adding `.Transform` to the common `GraphicsOptions` is a good idea unless we want to make **all** the processors in the core library to support it. Otherwise it would be confusing to users (there is a `Transform` property but it's not being applied ...).

This would mean that for example `.Resize(...)` also applies an arbitrary free transform, and `.Transform(...)` has an additional transform etc. This is not the way our API-s were evolving in the core library so far, and it's just too big change to even consider at this point IMO.

We should rather do an API simplification right within `ImageSharp.Drawing`.

`ShapeGraphicsOptions`, `TextGraphicsOptions`, `ShapeOptions`, `TextOptions` ... it's just too confusing. 

How about something like this:

```diff
- public class ShapeGraphicsOptions { ... }
- public class TextGraphicsOptions { ... }
+public class DrawingOptions
+{
+    public ShapeOptions ShapeOptions { get; set; }
+    public TextOptions ShapeOptions { get; set; }
+    public Matrix3x2 Transform { get; set; }
+    public GraphicsOptions GraphicsOptions { get; set; }
+}
```

Then instead of doing `.SetShapeOptions(x => x.Transform = ...)` or `.SetShapeOptions(x => x.Transform = ...)`, we can introduce a `SetTransform(Matrix3x2)` extension method.

src/ImageSharp.Drawing/Processing/Processors/Drawing/DrawPathProcessor.cs
tocsoft: we don't need to do this as the `Transform()` method does this internally.

src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
tocsoft: yeah this is one of the reasons its still draft, I had glyphs cut off with out it, its will be due to the size requirements being different when the item is rotated... i.e if its 10 x20 but rotated 90deg then we will actually need a 20x10 buffer (but add on the fact that any transform is possible and it complicated matters somewhat), i need to see if there is a more elegant(optimal solution to this one)

JimBobSquarePants: This should help there I think.

https://github.com/SixLabors/ImageSharp/blob/08b0320538636d369b50fcfce6e3fb2b8404a8d9/src/ImageSharp/Processing/Processors/Transforms/TransformUtils.cs#L323

src/ImageSharp.Drawing/Processing/Processors/Drawing/DrawPathProcessor.cs
antonfirsov: Cool!

src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
antonfirsov: If `path` is already transformed, shouldn't be `path.Bounds` be updated accordingly?

JimBobSquarePants: I would say it should. How we enforce that I do not yet know since each implementation sets that property.

src/ImageSharp.Drawing/Processing/DrawingTransformDefaultsExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace SixLabors.ImageSharp.Drawing.Processing
+{
+    /// <summary>
+    /// Adds extensions that allow the processing of images to the <see cref="Image{TPixel}"/> type.
+    /// </summary>
+    public static class DrawingTransformDefaultsExtensions
+    {
+        /// <summary>
+        /// Sets the default shape processing options against the image processing context.
```
antonfirsov, line 15:
```suggestion
        /// Sets the 2D transformation matrix to be used during rasterization when drawing shapes or text.
```

src/ImageSharp.Drawing/Processing/DrawingTransformDefaultsExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        /// <param name="context">The image processing context to store default against.</param>
+        /// <param name="matrix">The default matrix to use.</param>
```
antonfirsov, line 18:
```suggestion
        /// <param name="matrix">The matrix to use.</param>
```

src/ImageSharp.Drawing/Processing/DrawingTransformDefaultsExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <returns>The passed in <paramref name="context"/> to allow chaining.</returns>
+        public static IImageProcessingContext SetDrawingTransform(this IImageProcessingContext context, Matrix3x2 matrix)
+        {
+            context.Properties[typeof(DrawingTransformDefaultsExtensions)] = matrix;
+            return context;
+        }
+
+        /// <summary>
+        /// Sets the default shape processing options against the configuration.
```
antonfirsov, line 27:
```suggestion
        /// Sets the default 2D transformation matrix to be used during rasterization when drawing shapes or text.
```

src/ImageSharp.Drawing/Processing/DrawingTransformDefaultsExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        /// <param name="configuration">The configuration to store default against.</param>
+        /// <param name="matrix">The default matrix to use.</param>
+        public static void SetDrawingTransform(this Configuration configuration, Matrix3x2 matrix)
+        {
+            configuration.Properties[typeof(DrawingTransformDefaultsExtensions)] = matrix;
+        }
+
+        /// <summary>
+        /// Gets the default matrix to apply during raterization against the image processing context.
```
antonfirsov, line 37:
```suggestion
        /// Gets the default 2D transformation matrix to be used during rasterization when drawing shapes or text.
```

src/ImageSharp.Drawing/Processing/DrawingTransformDefaultsExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            var matrix = context.Configuration.GetDrawingTransform();
+
+            // do not cache the fall back to config into the the processing context
+            // in case someone want to change the value on the config and expects it re trflow thru
+            return matrix;
+        }
+
+        /// <summary>
+        /// Gets the default shape processing options against the image processing context.
```
antonfirsov, line 56:
```suggestion
        /// Gets the default 2D transformation matrix to be used during rasterization when drawing shapes or text.
```

src/ImageSharp.Drawing/Processing/DrawingTransformDefaultsExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        public static void SetDrawingTransform(this Configuration configuration, Matrix3x2 matrix)
+        {
+            configuration.Properties[typeof(DrawingTransformDefaultsExtensions)] = matrix;
+        }
+
+        /// <summary>
+        /// Gets the default matrix to apply during raterization against the image processing context.
+        /// </summary>
+        /// <param name="context">The image processing context to retrieve defaults from.</param>
+        /// <returns>The globaly configured default options.</returns>
```
antonfirsov, line 40:
```suggestion
        /// <returns>The matrix.</returns>
```

src/ImageSharp.Drawing/Processing/DrawingTransformDefaultsExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+            // do not cache the fall back to config into the the processing context
+            // in case someone want to change the value on the config and expects it re trflow thru
+            return matrix;
+        }
+
+        /// <summary>
+        /// Gets the default shape processing options against the image processing context.
+        /// </summary>
+        /// <param name="configuration">The configuration to retrieve defaults from.</param>
+        /// <returns>The globaly configured default options.</returns>
```
antonfirsov, line 59:
```suggestion
        /// <returns>The globaly configured default matrix.</returns>
```

src/ImageSharp.Drawing/Processing/DrawingOptionsDefaultsExtensions.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using SixLabors.ImageSharp.Processing;
+
+namespace SixLabors.ImageSharp.Drawing.Processing
+{
+    /// <summary>
+    /// Adds extensions that allow the processing of images to the <see cref="Image{TPixel}"/> type.
```
antonfirsov, line 9:
```suggestion
    /// Adds extensions that help working with <see cref="DrawingOptions" />.
```

src/ImageSharp.Drawing/Processing/DrawingTransformDefaultsExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+
+using System.Numerics;
+using SixLabors.ImageSharp.Processing;
+
+namespace SixLabors.ImageSharp.Drawing.Processing
+{
+    /// <summary>
+    /// Adds extensions that allow the processing of images to the <see cref="Image{TPixel}"/> type.
+    /// </summary>
+    public static class DrawingTransformDefaultsExtensions
```
antonfirsov, line 12: Shouldn't we merge this into the other extensions class? Feels like we have too many of them.

antonfirsov (commented):
Couple of Docs bugfixes + suggestion to merge `DrawingTransformDefaultsExtensions` with `DrawingOptionsDefaultsExtensions`.

src/ImageSharp.Drawing/Processing/DrawingTransformDefaultsExtensions.cs
tocsoft: sorry been busy and forgot finish up this.... merged the 2 new extension classes

antonfirsov (approved):
LGTM. Pushed a few more tests, optimizations and analyzer suggestions.

PR #128, opened 2021-04-19
More efficient BrushApplicators

### LINKED ISSUE #103: IPath: Remove unused features to enable simplifying and optimizing InternalPath
After merging #96, `InternalPath` will become unnecessarily heavy both code and memory-wise.

Ideally it should only contain a `PointF[]` to be fed to `TessellatedMultipolygon`, the following stuff seems to be heavy and unnecessary:

https://github.com/SixLabors/ImageSharp.Drawing/blob/acf438bb7bd420a6dc349ebd1d076a1447097a6a/src/ImageSharp.Drawing/Shapes/InternalPath.cs#L933-L965

The following features seem to rely on this data, I suggest to remove them from the public API, since they do not really serve drawing concerns:
- `IPath.FindIntersections`, `IPath.MaxIntersections`
- `IPath.Distance`
- `IPath.PointAlongPath`
- `IPath.Length`

Some of the above could be replaced with slow variants if we really want/need them.

### LINKED ISSUE #117: PathGradientBrush fails to render a complex polygon
SixLabors.ImageSharp v1.0.2
SixLabors.ImageSharp.Drawing v1.0.0-beta11
Running in Visual Studio 2019 v.16.8.3 on Windows 10 v.2004 in .NET 5.0.1-servicing.20575.16

Using the SixLabors.ImageSharp.Drawing.Processing.FillPathExtensions's Fill method, I can render a complex polygon using a SolidBrush, but using a PathGradientBrush the program doesn't finish executing and 'hangs' with high memory usage.

Code to reproduce:

```
        private static void Main(string[] args)
        {
            static void GradientPolygon()
            {
                static void DoComplexPolygon(IImageProcessingContext g, RectangleF rect)
                {
                    static PointF PointFromCircle(RectangleF rectBounds, double degrees)
                    {
                        static void RectCenter(RectangleF r, out float x, out float y)
                        {
                            x = (r.Left + r.Right) / 2.0f;
                            y = (r.Top + r.Bottom) / 2.0f;
                        }

                        RectCenter(rectBounds, out var xCenter, out var yCenter);
                        var radians = (Math.PI * degrees) / 180.0f;
                        var radiusH = rectBounds.Width / 2.0f;
                        var radiusV = rectBounds.Height / 2.0f;

                        var tmpX = xCenter + radiusH * (float)Math.Cos(radians);
                        var tmpY = yCenter - radiusV * (float)Math.Sin(radians);

                        return new PointF(tmpX, tmpY);
                    }

                    var iPoints = new List<PointF>();

                    for (double i = 0; i <= 360; i += 0.1)
                    {
                        iPoints.Add(PointFromCircle(rect, i));
                    }
                    var widthReduction = 50;

                    rect.Inflate(-widthReduction, -widthReduction);

                    for (double i = 360; i >= 0; i -= 0.1)
                    {
                        iPoints.Add(PointFromCircle(rect, i));
                    }

                    var path = new SixLabors.ImageSharp.Drawing.Polygon(new LinearLineSegment(iPoints.ToArray()));

                    //g.Fill(new SolidBrush(Color.Red), path); //this works
                    g.Fill(new PathGradientBrush(iPoints.ToArray(), new Color[] { Color.Red, Color.White, Color.Blue }), path); //this does not complete
                }

                var bitmap = new Image<Rgba32>(800, 600);

                bitmap.Mutate(x =>
                {
                    x.Clear(Color.White);
                    DoComplexPolygon(x, new RectangleF(137, 50, 526, 526));
                });

                var path = $@"{AppDomain.CurrentDomain.BaseDirectory}/Polygon{DateTime.Now.Ticks}.png";

                bitmap.SaveAsPng(path);

                new Process
                {
                    StartInfo = new ProcessStartInfo(path)
                    {
                        UseShellExecute = true
                    }
                }.Start();
            }

            GradientPolygon();
        }
    }
```

JimBobSquarePants:
### Prerequisites

### Description
Fix #117 

So far I've refactored the code to remove the `T[x, y]` contract and made `Apply(Span<float> scanline, int x, int y)` abstract. 

This allows use to reduce the per pixel pool request to a per-thread requests which should dramatically speed up performance. 

In order to do this I had to implement the internalizing and refactoring  of `InternalPath.FindIntersection` etc to remove the secondary per pixel request there to fix #103


src/ImageSharp.Drawing/Shapes/EllipsePolygon.cs
```diff
@@ -144,30 +141,17 @@ public IEnumerable<ISimplePath> Flatten()
...
-            Span<PointF> subBuffer = buffer.AsSpan(offset);
-            return this.innerPath.FindIntersections(start, end, subBuffer, intersectionRule);
-        }
-
-        /// <inheritdoc/>
-        int IPath.FindIntersections(PointF start, PointF end, Span<PointF> buffer, IntersectionRule intersectionRule)
-        {
-            return this.innerPath.FindIntersections(start, end, buffer, intersectionRule);
-        }
+        int IPath.FindIntersections(PointF start, PointF end, Span<PointF> intersections, Span<PointOrientation> orientations)
```
JimBobSquarePants, line 144: I don't know why these are hidden in this polygon but visible in others?

src/ImageSharp.Drawing/Shapes/RectangularPolygon.cs
```diff
@@ -467,36 +454,20 @@ public IEnumerable<ISimplePath> Flatten()
...
                 this.Height == other.Height &&
                 this.Width == other.Width;
-        }
 
-        /// <summary>
-        /// Equality comparer for two RectangularPolygons
-        /// </summary>
-        /// <param name="obj">The polygon to compare to.</param>
-        /// <returns>Returns a value indicating if the rectangles are equal.</returns>
+        /// <inheritdoc/>
```
JimBobSquarePants, line 463: This equality stuff is odd.

src/ImageSharp.Drawing/Shapes/PointOrientation.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp.Drawing
+{
+    /// <summary>
+    /// Represents the orientation of a point from a line.
+    /// </summary>
+    public enum PointOrientation
```
JimBobSquarePants, line 9: I had to make this public for the IPath API.

src/ImageSharp.Drawing/Shapes/InternalPath.cs
```diff
@@ -251,19 +239,19 @@ public int FindIntersectionsWithOrientation(PointF start, PointF end, Span<Point
...
             {
-                polyCorners -= 1;
+                polyCorners--;
             }
 
             int position = 0;
             Vector2 lastPoint = MaxVector;
 
+            // TODO: This request can happen in huge numbers.
             PassPointData[] precaclulate = ArrayPool<PassPointData>.Shared.Rent(this.points.Length);
```
JimBobSquarePants, line 249: Maybe we can check the length and stackalloc here?

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
```diff
@@ -310,14 +314,54 @@ private class PathGradientBrushApplicator<TPixel> : BrushApplicator<TPixel>
...
 
-            private (Edge edge, Intersection? info) FindIntersection(PointF start, PointF end)
+            /// <inheritdoc />
+            public override void Apply(Span<float> scanline, int x, int y)
+            {
+                MemoryAllocator memoryAllocator = this.Configuration.MemoryAllocator;
+                using IMemoryOwner<float> amountBuffer = memoryAllocator.Allocate<float>(scanline.Length);
+                using IMemoryOwner<TPixel> overlayBuffer = memoryAllocator.Allocate<TPixel>(scanline.Length);
+                using IMemoryOwner<PointF> intersectionsBuffer = memoryAllocator.Allocate<PointF>(this.maxIntersections);
+                using IMemoryOwner<PointOrientation> orientationsBuffer = memoryAllocator.Allocate<PointOrientation>(this.maxIntersections);
```
antonfirsov, line 324: Can we try the following:
- Move these to properties of a nested class `ThreadContextData`, then store it as a member `ThreadLocal<ThreadContextData>` allocated in the constructor
- The value can fetched accessed here in `Apply` from `this.threadContext.Value`
- `BrushApplicator.Dispose` should iterate through `this.threadContext.Values` disposing each

src/ImageSharp.Drawing/Shapes/Path.cs
```diff
@@ -159,28 +157,17 @@ public IEnumerable<ISimplePath> Flatten()
...
-        {
-            return this.InnerPath.FindIntersections(start, end, buffer);
-        }
-
-        /// <inheritdoc />
-        public int FindIntersections(PointF start, PointF end, PointF[] buffer, int offset, IntersectionRule intersectionRule)
-        {
-            return this.InnerPath.FindIntersections(start, end, buffer.AsSpan(offset), intersectionRule);
-        }
+        public int FindIntersections(PointF start, PointF end, Span<PointF> intersections, Span<PointOrientation> orientations)
```
antonfirsov, line 160: We have to internalize these as per #103. (This is very ugly as a public API contract exposing implementation details, normally we would iterate on the API design a lot here!)

Does any call site actually use `FindIntersections` through `IPath`, not knowing the subtype? If yes, simply cast the objects to `IInternalPathOwner` and delegate there to `InnerPath`.

src/ImageSharp.Drawing/Shapes/PointOrientation.cs
antonfirsov: We shouldn't. See previous comment.

src/ImageSharp.Drawing/Shapes/RectangularPolygon.cs
antonfirsov: Is there any code actually depending on it?

src/ImageSharp.Drawing/Shapes/InternalPath.cs
antonfirsov: I wouldn't bother. This code has to be simplified a lot (again #103). `FindIntersectionsWithOrientation` has to be dumped entirely if I'm not mistaken. Is there any call-site actually using it?

src/ImageSharp.Drawing/Shapes/PointOrientation.cs
JimBobSquarePants: So... Just to clarify. 

All polygon implementations (except `EllipsePolygon`) currently publicly implements `IPath.FindIntersections` which requires this type to be exposed publicly in order to remove the per pixel pooled allocation. 

Are you saying that the intersections methods should be moved to the `IInternalPathOwner` interface?

src/ImageSharp.Drawing/Shapes/RectangularPolygon.cs
JimBobSquarePants: I'll review and delete it if not. 

I think we should have a good look at our usage of structs/classes for types to make sure we're doing the right thing.

src/ImageSharp.Drawing/Shapes/PointOrientation.cs
antonfirsov: All `FindIntersections` overloads should be removed from `IPath` together with `IPath.MaxIntersections`, `IPath.Distance`, 
 `IPath.PointAlongPath`, `IPath.Length` and turned into implementation details. The way we achieve it, does not really matter.
 
If there are internal calls through `IPath`, we can just replace them with `((IInternalPathOwner)path).InternalPath.FindIntersections(...)`. If there are call sites where we know the specific polygon type, we can use `SpecificPolygonSubclass.FindIntersections(...)` after making it internal.

src/ImageSharp.Drawing/Shapes/InternalPath.cs
JimBobSquarePants: It's called directly by `InternalPath` and `ComplexPolygon`.

src/ImageSharp.Drawing/Shapes/PointOrientation.cs
JimBobSquarePants: The constructor for `ComplexPolygon` is thwarting me currently. It calculates the max intersections based on the sum of the property on given `IPath` collection.

JimBobSquarePants: I've restricted the constructor to `Path` since it's currently only used by `Polygon` and `Path`. Need to have a think about that....

antonfirsov: Good enough for now, we can figure out later how to handle this.

JimBobSquarePants: Found a use in the tests.

``` c#
new ComplexPolygon(new EllipsePolygon(new Vector2(603), 161f), new EllipsePolygon(new Vector2(603), 61f))
```

src/ImageSharp.Drawing/Shapes/ComplexPolygon.cs
```diff
@@ -111,34 +122,32 @@ public ComplexPolygon(params IPath[] paths)
...
-                PointInfo d = shape.Distance(point);
+                PointInfo d;
+                if (shape is IPathInternals internals)
+                {
+                    d = internals.Distance(point);
+                }
+                else
+                {
+                    // TODO: How can I solve this?
+                    throw new NotImplementedException();
```
JimBobSquarePants, line 149: @antonfirsov I'll need your help here. I don't know enough to implement it.

antonfirsov: You would have to call `path.Flatten().Points` and go through each `point[i] -> point[i+1]` line with `InernalPath.CalculateShorterDistance` ...

I don't see why we need `Distance` at all. The only use cases I see are: (1) tests (2) implementations of another `Distance` method. A perfect time to delete code I'd say.

JimBobSquarePants: Hurrah, done and thanks!

src/ImageSharp.Drawing/Shapes/Path.cs
JimBobSquarePants: I've internalized everything now.

JimBobSquarePants:
Ok... Now we're cooking. 

I haven't implemented [`ThreadLocal<T>` ](https://github.com/SixLabors/ImageSharp.Drawing/pull/128#discussion_r616476221)yet but have done a massive cleanup of the polygon code to internalize everything mentioned in #103 deleting unused properties `IPath.Length` and `IPath.Distance`.

I've also normalized all the visibility modifiers for `IPath`/`ISimplePath` and fixed documentation. Code coverage is way up compared to previously.

JimBobSquarePants:
OK. That's the `ThreadLocal<T>` approach implemented on all `BrushApplicator<T>` types so I'm marking this a s ready to review.

The following code derived from the example in #117 that generates 720 paths runs in around ~30 seconds on my machine now. I'd like to see how it runs on another. @tannergooding if you are able that would be awesome. 

```c#
private void GradientPolygonWith720Paths()
{
    static void DoComplexPolygon(IImageProcessingContext g, RectangleF rect)
    {
        static PointF PointFromCircle(RectangleF rectBounds, double degrees)
        {
            static void RectCenter(RectangleF r, out float x, out float y)
            {
                x = (r.Left + r.Right) / 2.0f;
                y = (r.Top + r.Bottom) / 2.0f;
            }

            RectCenter(rectBounds, out float xCenter, out float yCenter);
            double radians = Math.PI * degrees / 180.0f;
            float radiusH = rectBounds.Width / 2.0f;
            float radiusV = rectBounds.Height / 2.0f;

            float tmpX = xCenter + (radiusH * (float)Math.Cos(radians));
            float tmpY = yCenter - (radiusV * (float)Math.Sin(radians));

            return new PointF(tmpX, tmpY);
        }

        var iPoints = new List<PointF>();

        for (double i = 0; i <= 360; ++i)
        {
            iPoints.Add(PointFromCircle(rect, i));
        }

        const int widthReduction = 50;

        rect.Inflate(-widthReduction, -widthReduction);

        for (double i = 360; i >= 0; --i)
        {
            iPoints.Add(PointFromCircle(rect, i));
        }

        var path = new Polygon(new LinearLineSegment(iPoints.ToArray()));
        g.Fill(new PathGradientBrush(iPoints.ToArray(), new Color[] { Color.Red, Color.White, Color.Blue }), path);
    }

    using var bitmap = new Image<Rgba32>(800, 600);
    bitmap.Mutate(x =>
    {
        x.Clear(Color.White);
        DoComplexPolygon(x, new RectangleF(137, 50, 526, 526));
    });
}
```

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
JimBobSquarePants: Done

tannergooding:
On my box (AMD Ryzen 5950X), the sample above goes from `00:01:17.5690030` (approx 1 minute 18 seconds) to `00:00:12.8436367` (approx 13 seconds).

A profiling shows that the majority of time is now in `FindIntersectionsWithOrientation`, as expected:
image

src/ImageSharp.Drawing/Shapes/InternalPath.cs
```diff
@@ -139,109 +117,78 @@ internal enum PointOrientation
...
-        public int FindIntersectionsWithOrientation(PointF start, PointF end, Span<PointF> buffer, Span<PointOrientation> orientationsSpan)
+        public int FindIntersectionsWithOrientation(
+            PointF start,
+            PointF end,
+            Span<PointF> intersections,
+            Span<PointOrientation> orientations)
         {
+            DebugGuard.IsTrue(intersections.Length == orientations.Length, nameof(orientations), "Intersection and orientation lengths must match.");
+
             if (this.points.Length < 2)
```
tannergooding, line 186: This check is fairly hot, taking about 13% of the average method time.

It might be worth refactoring to be something like the following so the early exit check can be inlined:
```csharp
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public int FindIntersectionsWithOrientation(
            PointF start,
            PointF end,
            Span<PointF> intersections,
            Span<PointOrientation> orientations)
        {
            DebugGuard.IsTrue(intersections.Length == orientations.Length, nameof(orientations), "Intersection and orientation lengths must match.");

            if (this.points.Length < 2)
            {
                return 0;
            }
            
            return FindIntersectionsWithOrientationInternal(start, end, buffer, orientationsSpan);
        }
```

src/ImageSharp.Drawing/Shapes/InternalPath.cs
```diff
@@ -251,171 +198,176 @@ public int FindIntersectionsWithOrientation(PointF start, PointF end, Span<Point
...
-            PassPointData[] precaclulate = ArrayPool<PassPointData>.Shared.Rent(this.points.Length);
+            // Avoid pool overhead for short runs.
+            // This method can be called in high volume.
+            int pointsLength = this.points.Length;
+            int maxStackSize = 1024 / Unsafe.SizeOf<PassPointData>();
+            PassPointData[] rentedFromPool = null;
+            Span<PassPointData> buffer =
+                pointsLength > maxStackSize
+                ? (rentedFromPool = ArrayPool<PassPointData>.Shared.Rent(pointsLength))
+                : stackalloc PassPointData[maxStackSize];
```
tannergooding, line 215: This is the other "hot" part of the method, taking on average 22% of the method time.

src/ImageSharp.Drawing/Shapes/InternalPath.cs
```diff
@@ -251,171 +198,176 @@ public int FindIntersectionsWithOrientation(PointF start, PointF end, Span<Point
...
+
+            Span<PassPointData> precalculate = buffer.Slice(0, pointsLength);
+
+            // Pre calculate relative orientations X places ahead and behind
+            Vector2 startToEnd = end - start;
+            PointOrientation prevOrientation = CalulateOrientation(startToEnd, this.points[polyCorners - 1].Point - end);
```
tannergooding, line 221: This call also ends up fairly "hot", taking about 11% of the average method time. I haven't looked why yet, but I'd expect its either due to copying or bounds checks.

tannergooding: To clarify, the profile shows this line in particular and not the `nextOrientation = ...` or `nextPlus1Orientation = ...` lines as being expensive.

src/ImageSharp.Drawing/Shapes/InternalPath.cs
antonfirsov: Yay, looks like a low hanging fruit!

antonfirsov:
@JimBobSquarePants FYI `FindIntersections` is currently only used by `PathGradientBrush`, the code is highly suboptimal and full of - now - dummy stuff inherited from the days it has been used to render polygons. For example note that `orientations` buffer is always filled with defaults, noone ever writes to it.

src/ImageSharp.Drawing/Shapes/InternalPath.cs
JimBobSquarePants: Ace, yep can easily do that. Thanks!

src/ImageSharp.Drawing/Shapes/InternalPath.cs
JimBobSquarePants: This one could be more tricky. The problem is that I don't have the domain knowledge working with polygons.

JimBobSquarePants:
>@JimBobSquarePants FYI FindIntersections is currently only used by PathGradientBrush, the code is highly suboptimal and full of - now - dummy stuff inherited from the days it has been used to render polygons. For example note that orientations buffer is always filled with defaults, noone ever writes to it.

@antonfirsov is this something we can migrate to use your PolygonScanner then? Would be nice to be able to delete all this.

antonfirsov:
@JimBobSquarePants no, because PolygonScanner is a very specific code for for intersecting a polygon with a huge number of sub-pixel vertical lines.

If I get the code right, PathGradientBrush needs to find the closest intersection of the path to `center` on the line `(x,y) -> center`:

https://github.com/SixLabors/ImageSharp.Drawing/blob/34ab238951f180d11abc4687d9fde9acebec9acd/src/ImageSharp.Drawing/Processing/PathGradientBrush.cs#L309-L312

`InternalPath.FindIntersections` is quite a big sledghummer to achieve this, the whole code together with all the allocations we optimized in the PR is actually unnecessary.

Using the `IInternalPath` abstraction through `Edge.Path` is also unnecessary, it is just a `LinearLineSegment` of 2 points:

https://github.com/SixLabors/ImageSharp.Drawing/blob/34ab238951f180d11abc4687d9fde9acebec9acd/src/ImageSharp.Drawing/Processing/PathGradientBrush.cs#L57-L69

One can just iterate through the edges with a dumb method [like this](https://stackoverflow.com/a/1968345) + a minimum search to get that intersection point.

JimBobSquarePants:
@antonfirsov Yeah... That's currently beyond me. I'll need someone else to implement it.

src/ImageSharp.Drawing/Shapes/IPathInternals.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <param name="end">The end position.</param>
+        /// <param name="intersections">The buffer for storing each intersection.</param>
+        /// <param name="orientations">
+        /// The buffer for storing the orientation of each intersection.
+        /// Must be the same length as <paramref name="intersections"/>.
+        /// </param>
+        /// <returns>
+        /// The number of intersections found.
+        /// </returns>
+        int FindIntersections(PointF start, PointF end, Span<PointF> intersections, Span<PointOrientation> orientations);
```
antonfirsov, line 32: Here is the only use case of the method above:

https://github.com/SixLabors/ImageSharp.Drawing/blob/34ab238951f180d11abc4687d9fde9acebec9acd/src/ImageSharp.Drawing/Processing/PathGradientBrush.cs#L170-L174

It means we only use `Path.FindIntersections`, the rest is unused code. I would make `FindIntersections` an internal method on `Path` and delete all the other implementations.

src/ImageSharp.Drawing/Shapes/IPathInternals.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <param name="intersections">The buffer for storing each intersection.</param>
+        /// <param name="orientations">
+        /// The buffer for storing the orientation of each intersection.
+        /// Must be the same length as <paramref name="intersections"/>.
+        /// </param>
+        /// <param name="intersectionRule">How intersections should be handled.</param>
+        /// <returns>
+        /// The number of intersections found.
+        /// </returns>
+        int FindIntersections(
```
antonfirsov, line 49: If we act on the previous comment, this can be deleted as well together with all implementations.

src/ImageSharp.Drawing/Shapes/ISimplePath.cs
```diff
@@ -1,13 +1,12 @@
...
 // Licensed under the Apache License, Version 2.0.
 
 using System;
-using System.Collections.Generic;
 
 namespace SixLabors.ImageSharp.Drawing
 {
     /// <summary>
-    /// Represents a logic path that can be drawn
+    /// Represents a simplified logic path that can be drawn.
```
antonfirsov, line 9:
```suggestion
    /// Represents a simple (non-composite) path defined by a series of points.
```

src/ImageSharp.Drawing/Shapes/PathExtensions.Outline.cs
```diff
@@ -10,10 +10,10 @@
 
 namespace SixLabors.ImageSharp.Drawing
 {
-    /// <summary>
+    /// <content>
     /// Path extensions to generate outlines of paths.
-    /// </summary>
-    public static class Outliner
+    /// </content>
+    public static partial class PathExtensions
```
antonfirsov, line 16: `OutlineExtensions` or `OutlinePathExtensions` ?

src/ImageSharp.Drawing/Shapes/PointOrientation.cs
antonfirsov, line 9: Left public by mistake I assume.
```suggestion
    internal enum PointOrientation
```

tests/ImageSharp.Drawing.Tests/Drawing/Paths/ClearRectangle.cs
```diff
@@ -23,7 +24,7 @@ public void Brush()
             FillPathProcessor processor = this.Verify<FillPathProcessor>();
 
             Assert.NotEqual(this.shapeOptions, processor.Options.ShapeOptions);
-            Assert.Equal(this.RectanglePolygon, processor.Shape);
+            Assert.True(RectangularPolygonValueComparer.Equals(this.RectanglePolygon, processor.Shape));
```
antonfirsov, line 27: Nit:
```C#
Assert.Equal(this.RectanglePolygon, processor.Shape, RectangularPolygonValueComparer.Instance);
```

antonfirsov (commented):
There is trivial opportunity to delete a lot of code, and simplify stuff further, but it means some chore work. I'm fine doing it in a separate PR if you feel it's too much on you.

src/ImageSharp.Drawing/Shapes/PathExtensions.Outline.cs
JimBobSquarePants: I would prefer to keep a single partial but forgot we use separate class names for the main library. Will update.

src/ImageSharp.Drawing/Shapes/PointOrientation.cs
JimBobSquarePants: Yup!

tests/ImageSharp.Drawing.Tests/Drawing/Paths/ClearRectangle.cs
JimBobSquarePants: I'll have to make it an `IValueComparer` instance

src/ImageSharp.Drawing/Shapes/IPathInternals.cs
JimBobSquarePants: Done.... wish I'd spotted on `Path` was being used. Took me ages to hack a workaround! 😄

src/ImageSharp.Drawing/Shapes/PathExtensions.Outline.cs
JimBobSquarePants: Done, same with the clipping extensions

src/ImageSharp.Drawing/Shapes/IPathInternals.cs
JimBobSquarePants: Done!

src/ImageSharp.Drawing/Processing/PatternBrush.cs
```diff
@@ -145,29 +152,74 @@ private class PatternBrushApplicator<TPixel> : BrushApplicator<TPixel>
...
-                        destinationRow,
-                        overlaySpan,
-                        amountSpan);
+                    this.threadContextData.Dispose();
+                }
+
+                this.isDisposed = true;
+            }
+
+            private sealed class ThreadContextData : IDisposable
```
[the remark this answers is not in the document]
JimBobSquarePants: Yep, agreed. I'd like to get the intersections code simplified first though so align the required allocation types.

JimBobSquarePants:
@antonfirsov deleted everything that was unused. Much happier now. I've left the final optimization for you since I don't know how plus I'm unwell again and really should be resting.

antonfirsov:
@JimBobSquarePants will take a final look tomorrow, get better!

antonfirsov (approved):
LGTM, great work, will follow up with some stuff if everything goes well!

Issue #137, opened 2021-05-25
Beta12 nuget package lib for netcoreapp3.1 target has incorrect version of SixLabors.ImageSharp.Drawing.dll


starkna:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp.Drawing
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp.Drawing/issues) and [closed](https://github.com/SixLabors/ImageSharp.Drawing/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
When referencing the beta12 nuget package for netcoreapp3.1 target, the wrong version of the SixLabors.ImageSharp.Drawing.dll is included (0.0.0-alpha.0+39acb53427a-4b0d0f989c...) instead of the correct 1.0.0-beta12 version.

### Steps to Reproduce
Add nuget package reference for Sixlabors.ImageSharp.Drawing beta 12 to netcoreapp3.1 (or net5.0) project).

### System Configuration
N/A

- ImageSharp.Drawing version:  beta12
- Other ImageSharp packages and versions:  N/A
- Environment (Operating system, version and so on):  Windows 10 20H2
- .NET Framework version: Net5.0
- Additional information:

JimBobSquarePants:
@starkna thanks for raising this! 

I've confirmed there's definitely an issue and it's only affecting the Core3.1 build. I'll have a look at our build process and if I can identify a fix will release a new beta today.

![image](https://user-images.githubusercontent.com/385879/119511058-f8e67280-bd69-11eb-9ad2-a1ff5685986e.png)

starkna:
@JimBobSquarePants Thanks for the quick reply - appreciate it!

JimBobSquarePants:
It's definitely an odd one and doesn't seem to affect any of our other repos with identical setup. I'm experimenting with a simplified build process to see if I can narrow a cause down.

JimBobSquarePants:
Nope. I'm stumped. There's no justification for this I can find.

https://github.com/SixLabors/ImageSharp.Drawing/runs/2667132314?check_suite_focus=true#step:5:15

JimBobSquarePants:
I've asked in the MinVer repo
https://github.com/adamralph/minver/discussions/518

JimBobSquarePants:
[...]

PR #138, opened 2021-05-25
Fix Build Process.

JimBobSquarePants:
### Prerequisites

### Description

There's something funny going on with versioning for .NET Core 3.1. See #137 

I've checked and none of our other repositories are affected so I'm assuming (hoping) it's something to do with this repositories git history

Looking at the publish step there's a warning that the project directory is not a valid git folder. I'm going to see if simplifying the process to use standard action options fixes the issue.
https://github.com/SixLabors/ImageSharp.Drawing/runs/2644202851?check_suite_focus=true#step:5:82

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -52,7 +52,7 @@ jobs:
             - name: Git Checkout
               uses: actions/checkout@v2
               with:
-                fetch-depth: 2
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Ok... Versioning on build is correct but the only way to really test this (_since I couldn't replicate it locally_) is to merge to master and let publish run.

PR #144, opened 2021-06-18
Implenting addArc feature

derBobo:
### Prerequisites

### Description
This should implement the requested features from #4


derBobo:
Currently the transformation feature is still not completed, I think there are two ways to implement this:
1. (current code tries this) Calculate everything, annoying because for every angle for two points on the circle there are at least 2 possibilities (clockwise and anticlockwise) .
2. Every elliptical arc stores it current transformation matrix and uses it to calculate the current points

antonfirsov:
@derBobo feel free to go with 2., we can always optimize later, the main goal is to close the feature gap.

derBobo:
Has anyone any idea, why this code always returns the point (0|0)? It works fine without the usage of PointF.Transform
```
 private PointF CalculatePoint(float angle)
        {
            float x = (this.firstRadius * MathF.Sin(MathF.PI * angle / 180) * MathF.Cos(MathF.PI * this.rotation / 180)) - (this.secondRadius * MathF.Cos(MathF.PI * angle / 180) * MathF.Sin(MathF.PI * this.rotation / 180)) + this.center.X;
            float y = (this.firstRadius * MathF.Sin(MathF.PI * angle / 180) * MathF.Sin(MathF.PI * this.rotation / 180)) + (this.secondRadius * MathF.Cos(MathF.PI * angle / 180) * MathF.Cos(MathF.PI * this.rotation / 180)) + this.center.Y;
            return PointF.Transform(new PointF(x, y), this.transformation);
        }
```

derBobo:
Should I also change the EllipsePlygon class to use the EllipticalArcLineSegment instead of CubicBezierLineSegment?

JimBobSquarePants:
@derBobo i don’t actually know the answer to that. @tocsoft what do you think?

tocsoft:
from memory its really an internal implentation detail on what sort of line segment we are using so I would say unless it makes a noticable/more acurate/faster difference to the output then I personally wouldn't bother. But saying all that as long as we have sufficent test coverage then I can't imagine it would cause too much harm.

JimBobSquarePants:
Thanks @tocsoft makes sense!

derBobo:
does anyone have more ideas fore Xunit tests ?

tests/ImageSharp.Drawing.Tests/Shapes/EllipticalArcLineSegmentTest.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System.Collections.Generic;
+using System.Numerics;
+using Xunit;
+
+namespace SixLabors.ImageSharp.Drawing.Tests
+{
+    public class EllipticalArcLineSegmentTest
```
JimBobSquarePants, line 10: I would add additional tests for things like 0 radius values and sweepAngle <360>

JimBobSquarePants: In addition I would add tests testing the extension method itself. Actually draw the arc and create a reference output for future comparison.

src/ImageSharp.Drawing/Shapes/PathBuilder.cs
```diff
@@ -199,6 +199,23 @@ public PathBuilder AddBezier(PointF startPoint, PointF controlPoint1, PointF con
...
 
+        /// <summary>
+        /// Adds a elliptical arc to the current  figure
+        /// </summary>
+        /// <param name="center">center point of the arc</param>
+        /// <param name="firstRadius">first radius</param>
+        /// <param name="secondRadius">second radius</param>
+        /// <param name="rotation">rotation of the first radius to the x Axis in degree</param>
+        /// <param name="startAngle">starting angle of arc in degree</param>
+        /// <param name="sweepAngle">sweeping angel of the arc in degree</param>
```
JimBobSquarePants, line 210: I'm curious. These parameters are very different from the various options offered by System.Drawing what was the inspiration for them? 

I wonder if we should simply be replicating (or at least creating something similar) to those overloads?

https://docs.microsoft.com/en-us/dotnet/api/system.drawing.drawing2d.graphicspath.addarc?view=net-5.0

derBobo: Except of the rotation parameter they are similar to https://docs.microsoft.com/en-us/dotnet/api/system.drawing.drawing2d.graphicspath.addarc?view=net-5.0#System_Drawing_Drawing2D_GraphicsPath_AddArc_System_Single_System_Single_System_Single_System_Single_System_Single_System_Single_ maybe the order changed .
 The reason for the rotation parameter was my first attempt to implement the usage of transformation matrices, where I recognized, that by allowing to rotate the arc there are more possibilities for the the created path.

Replicating some of the overloads was actually on my to do list, but I hadn't the time to do that

JimBobSquarePants: Ah I see the rotation parameter is used in the [SVG arc implementation](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Paths#arcs). Thanks for clarifying.

Ok let's keep it then since it seems very useful like you say but rename it to `rotationX`. 
We should also name the radius parameters `radiusX` and `radiusY`. 

I think it would be beneficial to copy parameter descriptions for the intellisense docs from the Microsoft docs.

We'll need to add the overloads as it would be cumbersome to only have this one available as an option.

Thanks!

antonfirsov:
> does anyone have more ideas fore Xunit tests ?

We need tests that actually draw the `EllipticalArcLineSegment`.

Example tests doing some drawing of `IPath`, you can add the new ones here:
https://github.com/SixLabors/ImageSharp.Drawing/blob/master/tests/ImageSharp.Drawing.Tests/Drawing/DrawPathTests.cs

These tests are validating the output agains reference images found here:
https://github.com/SixLabors/ImageSharp.Drawing/tree/master/tests/Images/ReferenceOutput/Drawing/DrawPathTests

The images are Debug-Saved under the folder `/Images/ActualOutput/Drawing/DrawPathTests/`. If you think the output is correct, just copy the files to the appropriate `ReferenceOutput` folder.

tests/ImageSharp.Drawing.Tests/Shapes/EllipticalArcLineSegmentTest.cs
derBobo: > I would add additional tests for things like 0 radius values and sweepAngle <360>

0  valued radii are tested now and sweepAngle  360° is checked via the addition to DrawPathTests.cs

derBobo:
I think this feature now is ready for review.

JimBobSquarePants:
@SixLabors/core would someone have a chance to review this?

tocsoft (approved):
lgtm 👍 

@derBobo thanks for this, and your patience in getting this in.

derBobo:
You're welcome. Am 09.09.2021 16:13 schrieb Scott Williams ***@***.***>:

@tocsoft approved this pull request.



lgtm 👍

@derBobo thanks for this, and your patience in getting this in.

Issue #13, opened 2019-08-11
Shapes: Internal ClipperException thrown by `GenerateOutline`


antonfirsov:
```csharp
PointF naan = new PointF(float.NaN, float.NaN);
Polygon path = new Polygon(new LinearLineSegment(new[] { naan, naan, naan, naan }));

// Throws internal exception:
path.GenerateOutline(10);
```

We should not throw internal exceptions, which are not communicated on the API surface.
Probably worth to review the library for more cases like this.

antonfirsov:
Easy to fix, part of API cleanup, I believe we should do it for 1.0.

obayomy:
@antonfirsov  
I was working on this issue when I noticed your recent comment and don't know if it is still possible to PR. My change is to move the (ClipperOffset) calls to a wrapper where such cases could be checked and throws the (ClipperException), should I make a PR?

obayomy:
@antonfirsov  
I was working on this issue when I noticed your recent comment and don't know if it is still possible to PR. My change is to move the (ClipperOffset) calls to a wrapper where such cases could be checked and throws the (ClipperException), should I make a PR?

antonfirsov:
@obayomy yes please! We should translate `ClipperException` to an `InvalidOperationException`.

Issue #10, opened 2019-07-11
Add possibility to make intersection of shapes


ToniaDemchuk:
This library has only simple difference clipping. It would be nice to have other types of clipping (intersect, union). 

As I see, you use ClippingLib under the hood and it has support for intersection.

antonfirsov:
Makes sense, but low priority, moving to Future. Community PR-s are welcome.

JimBobSquarePants:
On it. See #265 265

PR #145, opened 2021-06-19
Added wrapper for ClipperOffset algorithm to fix internal exception

obayomy:
### Prerequisites

### Description

- Made a wrapper for `ClipperOffset` in the same manner of `Clipper` wrapper in `PolygonsClipper` 
- Added check for the NaN case in #13


antonfirsov (approved):
The issue described in #13 is that `ClipperException` is internal.

This is likely a mistake, since [`SixLabors.ImageSharp.Drawing.PolygonClipper.Clipper`](https://github.com/SixLabors/ImageSharp.Drawing/blob/c29e5de8bb1ed0afd11d71cb75ede5fd4bee0644/src/ImageSharp.Drawing/Shapes/PolygonClipper/Clipper.cs) is public. (Which I haven't noticed when opening that issue.)

The way to fix #13 is to make `ClipperException` public and document it in all methods that can possibly throw it. (All methods in `OutlinePathExtensions` or more).

The new `ClipperOffset` API needs justification, examples, an API review, and extensive test coverage before we can merge it.

obayomy:
Thank you for clarifying.

The current boundary between the `ImageSharp.Drawing.Shapes` and `ClipperLib` is a bit confusing to me as there are two different approaches for using `ClipperLib`: the `ClipPathExtensions` uses the external library through the `PolygonClipper.Clipper`  wrapper:
https://github.com/SixLabors/ImageSharp.Drawing/blob/423bf06a6b8bfce3a9991bede8af3e550c6742ca/src/ImageSharp.Drawing/Shapes/ClipPathExtensions.cs#L20-L30
While the `OutlinePathExtensions` uses the external library directly:
https://github.com/SixLabors/ImageSharp.Drawing/blob/423bf06a6b8bfce3a9991bede8af3e550c6742ca/src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs#L80-L88
I moved the `ClipperLib` calls in `OutlinePathExtensions` to the new `PolygonClipper.Clipper` because I think it should be beneficial for the following points:

1. Make the usage of `ClipperLib` consistent.
2. Help when wrapping the internal `ClipperLib.ClipperException`  to a public one, as the lines that makes direct calls to `ClipperLib` will be minimized  and located only in the API. 
3. Help when working with #10 as new functions will be exposed directly from the API

src/ImageSharp.Drawing/Shapes/PolygonClipper/ClipperOffset.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using ClipperLib;
+
+namespace SixLabors.ImageSharp.Drawing.Shapes.PolygonClipper
+{
+    /// <summary>
+    /// Wrapper for clipper offset
+    /// </summary>
+    public class ClipperOffset
```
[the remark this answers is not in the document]
obayomy: The current boundary between the `ImageSharp.Drawing.Shapes` and `ClipperLib` is a bit confusing to me as there are two different approaches for using `ClipperLib`: the `ClipPathExtensions` uses the external library through the `PolygonClipper.Clipper`  wrapper:
https://github.com/SixLabors/ImageSharp.Drawing/blob/423bf06a6b8bfce3a9991bede8af3e550c6742ca/src/ImageSharp.Drawing/Shapes/ClipPathExtensions.cs#L20-L30
While the `OutlinePathExtensions` uses the external library directly:
https://github.com/SixLabors/ImageSharp.Drawing/blob/423bf06a6b8bfce3a9991bede8af3e550c6742ca/src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs#L80-L88
I moved the `ClipperLib` calls in `OutlinePathExtensions` to the new `PolygonClipper.Clipper` because I think it should be beneficial for the following points:

1. Make the usage of `ClipperLib` consistent.
2. Help when wrapping the internal `ClipperLib.ClipperException`  to a public one, as the lines that makes direct calls to `ClipperLib` will be minimized  and located only in the API. 
3. Help when working with #10 as new functions will be exposed directly from the API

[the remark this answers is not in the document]
obayomy: An example for the second point: 
When translating the exception that can be raised when calling `ClipperLib.ClipperOffset.AddPath()`  to a public one, only this line would be translated:
```
  lock (this.syncRoot)
  {
      this.innerClipperOffest.AddPath(points, this.Convert(jointStyle), endCapStyle);
  }
```
While trying to translate in `OutlinePathExtensions` will require translating every line/method that uses the external library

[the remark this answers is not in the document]
obayomy: I only made the `PolygonsClipper.ClipperOffset` public to be in the same manner as `PolygonsClipper.Clipper`. However, I think both should be internal, and the 'PolygonsClipper.ClipperException' should be changed to public as you clarified.

src/ImageSharp.Drawing/Shapes/PolygonClipper/ClipperOffset.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <param name="pathPoints">The path points</param>
+        /// <param name="jointStyle">Joint Style</param>
+        /// <param name="endCapStyle">Endcap Style</param>
+        /// <exception cref="ClipperException">AddPath: Open paths have been disabled.</exception>
+        private void AddPath(ReadOnlySpan<PointF> pathPoints, JointStyle jointStyle, EndType endCapStyle)
+        {
+            var points = new List<IntPoint>();
+            foreach (PointF v in pathPoints)
+            {
+                if (float.IsNaN(v.X) || float.IsNaN(v.Y))
```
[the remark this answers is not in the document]
obayomy: Okay, I will remove it and translate the internal exception instead.

src/ImageSharp.Drawing/Shapes/PolygonClipper/ClipperOffset.cs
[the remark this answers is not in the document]
antonfirsov: Hm, this leads very far. To solve #10 in general, I would prefer to see and discuss an API proposal before going into implementation details.

I also feel like it would be better to keep `Clipper` internal, since a general solution for #10 might break that API, but I'm not sure about this, could use some help from @tocsoft to decide. Unfortunately, I lack the time to think about a proper API for this alone. (And it's not a high-priority thing to get 1.0 released.)

If  @tocsoft doesn't get the chance to react here, let's just turn `ClipperOffest` for now.

src/ImageSharp.Drawing/Shapes/PolygonClipper/ClipperOffset.cs
```diff
@@ -0,0 +1,146 @@
...
+        public ComplexPolygon Execute(float width)
+        {
+            var tree = new List<List<IntPoint>>();
+            lock (this.syncRoot)
+            {
+                try
+                {
+                    this.innerClipperOffest.Execute(ref tree, width * ScalingFactor / 2);
+                }
+                catch (ClipperLib.ClipperException exception)
```
JimBobSquarePants, line 43: Took me a while to get my head around this but I think I understand the intent here. 

We cannot, however wrap the exception in this manner since the stack trace will be lost. 

There are two ways to deal with this. 

1. Continue to use the `try...catch` pattern but wrap `InvalidOperationException` instead of `ClipperLib.ClipperException`, passing the type as an internal exception.
2. Use only a single public `ClipperException` everywhere and don't use `try... catch`.

I would go with option 2 since the first option seems like massive complexity where there doesn't need to be.

src/ImageSharp.Drawing/Shapes/PolygonClipper/ClipperOffset.cs
```diff
@@ -0,0 +1,146 @@
...
+        /// </summary>
+        /// <param name="path">The path.</param>
+        /// <param name="jointStyle">Joint Style</param>
+        /// <param name="endCapStyle">Endcap Style</param>
+        /// <exception cref="ClipperException">AddPath: Invalid Path</exception>
+        public void AddPath(IPath path, JointStyle jointStyle, EndCapStyle endCapStyle)
+        {
+            if (path is null)
+            {
+                throw new ArgumentNullException(nameof(path));
```
JimBobSquarePants, line 80: Use `Guard.NotNull(...)`

src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs
```diff
@@ -70,24 +71,18 @@ public static IPath GenerateOutline(this IPath path, float width, ReadOnlySpan<f
         /// <param name="jointStyle">The style to render the joints.</param>
         /// <param name="patternSectionCapStyle">The style to render between sections of the specified pattern.</param>
         /// <returns>A new path representing the outline.</returns>
+        /// <exception cref="ClipperException">Execute: Couldn't caculate Offset</exception>
```
JimBobSquarePants, line 74: `calculate`, here and elsewhere

JimBobSquarePants (changes requested):
More changes than I would have initially done but I see why you have made them and they are definitely beneficial. If we can simplify the exception handling then I'm happy to merge this. Thanks!

src/ImageSharp.Drawing/Shapes/PolygonClipper/ClipperOffset.cs
obayomy: Replaced all null checking with `Guard` in `ClipperOffset.cs `as well as `Clipper.cs`

src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs
obayomy: Replaced `Execute` with `Calculate` in all places

tocsoft: The issue wont have been the word Execute but the miss-spelling of `calculate`

```diff
-      /// <exception cref="ClipperException">Execute: Couldn't caculate Offset</exception>
+      /// <exception cref="ClipperException">Execute: Couldn't calculate Offset</exception>
```

src/ImageSharp.Drawing/Shapes/PolygonClipper/ClipperOffset.cs
obayomy: Applied the second option. Changed the access modifier of  `ClipperLib.ClipperException` in `clipper.cs` to `public`, and excluded the `ClipperException.cs` in `PolygonClipper` Folder to avoid conflict.

tests/ImageSharp.Drawing.Tests/Shapes/PolygonClipper/ClipperTests.cs
```diff
@@ -138,14 +138,13 @@ public void ClipperOffsetThrowsPublicException()
...
 
-            // Throws internal exception:
             try
             {
                 path.GenerateOutline(10);
             }
             catch (System.Exception ex)
             {
-                Assert.True(ex is ClipperException);
+                Assert.True(ex is ClipperLib.ClipperException);
```
JimBobSquarePants, line 147: We don't want the exception to be a nested type.

src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs
```diff
@@ -22,7 +22,7 @@ public static class OutlinePathExtensions
         /// <param name="width">The final width outline</param>
         /// <param name="pattern">The pattern made of multiples of the width.</param>
         /// <returns>A new path representing the outline.</returns>
-        /// <exception cref="ClipperException">Execute: Couldn't caculate Offset</exception>
+        /// <exception cref="ClipperException">Calculate: Couldn't caculate Offset</exception>
```
JimBobSquarePants, line 25: Sorry, I meant fix the spelling. `caculate` should be `calculate`

src/ImageSharp.Drawing/Shapes/PolygonClipper/ClipperOffset.cs
JimBobSquarePants: If the second type is unused don’t be afraid to delete it.

tests/ImageSharp.Drawing.Tests/Shapes/PolygonClipper/ClipperTests.cs
obayomy: `ClipperLib` is not a type, it's just the namespace of the clipper library.

src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs
obayomy: No problem. Sorry for missing the type, will fix it.

tests/ImageSharp.Drawing.Tests/Shapes/PolygonClipper/ClipperTests.cs
JimBobSquarePants: Ah right… can you please remove that namespace prefix then. Once you’ve removed the duplicate type there shouldn’t be any collisions. Thanks!

src/ImageSharp.Drawing/Shapes/PolygonClipper/ClipperOffset.cs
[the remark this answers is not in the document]
JimBobSquarePants: >I also feel like it would be better to keep Clipper internal
@antonfirsov I'm seconding that and making it internal for now.

JimBobSquarePants:
@obayomy I had some time so pushed the final tweaks to give you a hand. Thanks ever so much for your help! 👍

JimBobSquarePants (approved):

PR #149, opened 2021-07-07
Allow implementing custom brush types

### LINKED ISSUE #20: Allow implementing custom brush types
For now, `IBrush` is supposed to return an instance of `BrushApplicator`, but both its constructor and indexer are defined with an `internal` modifier, which makes it practically impossible to create a custom brush implementation.

It can prevent users from adopting ImageSharp in case they need a type of brush that the library doesn't support, as described in SixLabors/ImageSharp#969, for example.

JimBobSquarePants:
### Prerequisites

### Description

Fixes #20 allowing custom brushes inheriting `BrushApplicator`


antonfirsov (approved):
LGTM

Issue #94, opened 2020-10-16
Review and fix parameter names and XML docs in extension methods


antonfirsov:
There are many leftovers from the time when extension methods used to operate on an image:

https://github.com/SixLabors/ImageSharp.Drawing/blob/799d6afce0749f3316f6e766ac8884cccc9ef46b/src/ImageSharp.Drawing/Processing/Extensions/DrawLineExtensions.cs#L95-L100

antonfirsov:
I believe it would be better to address this in RC1.

JimBobSquarePants:
Agreed

JimBobSquarePants:
I've been reading through the docs and it looks like I've fixed all the issues over time in other PRs. I'm going to close this as it's looking good.

antonfirsov:
This is not true:

https://github.com/SixLabors/ImageSharp.Drawing/blob/b059f3142106b9bf12922acd5f25156a41cc1fe6/src/ImageSharp.Drawing/Processing/Extensions/DrawBezierExtensions.cs#L76-L82

An easy way to address this would be to run a find & replace for `The image this method extends`.

PR #150, opened 2021-07-10
Remove Region and add new Clip Extension

### LINKED ISSUE #105: Remove Region and implement "ProcessInsideShape" within ImageSharp.Drawing
We want to enable clipping interest region of arbitrary processors by an arbitrary polygon. If we agree that solving this problem with `Region` is not feasible, we need to remove that type for 1.0, and add the [`ProcessInsideShape` utility](https://github.com/SixLabors/Samples/blob/32557bf359066c07358572e7147e638608522503/ImageSharp/CustomImageProcessor/Program.cs#L43-L46) to the core library.

For more details, see the discussion starting with https://github.com/SixLabors/ImageSharp.Drawing/pull/96#discussion_r512502959.

JimBobSquarePants:
### Prerequisites

### Description

Fixes #105 

- Removes all references to `Region` and inheriting types.
- Adds new `Clip(IImageProcessingContext, IPath, Action<IImageProcessingContext>)` extension with functionality copied from [samples](https://github.com/SixLabors/Samples/blob/38faba5caf866c0168ea67a5e7c37c06d3043775/ImageSharp/CustomImageProcessor/Program.cs)


src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor{TPixel}.cs
```diff
@@ -325,6 +325,11 @@ public void EndGlyph()
                 {
                     IPath path = this.builder.Build();
 
+                    if (path.Bounds.Equals(RectangleF.Empty))
```
JimBobSquarePants, line 328: I discovered that this was throwing in `TessellatedMultipolygon.Create` during testing with the current fonts library.

JimBobSquarePants: I moved this fix to another PR and added tests

src/ImageSharp.Drawing/Processing/Extensions/FillPathExtensions.cs
```diff
@@ -7,42 +7,30 @@
...
-        /// <param name="source">The image this method extends.</param>
-        /// <param name="brush">The brush.</param>
-        /// <param name="path">The path.</param>
-        /// <returns>The <see cref="IImageProcessingContext"/> to allow chaining of operations.</returns>
-        public static IImageProcessingContext Fill(this IImageProcessingContext source, IBrush brush, IPath path) =>
-            source.Fill(source.GetDrawingOptions(), brush, path);
+            source.Fill(new SolidBrush(color), path);
 
         /// <summary>
         /// Flood fills the image in the shape of the provided polygon with the specified brush..
```
tocsoft, line 28:
```suggestion
        /// Flood fills the image in the shape of the provided polygon with the specified brush.
```

src/ImageSharp.Drawing/Processing/Extensions/FillPathExtensions.cs
```diff
@@ -7,42 +7,30 @@
...
 {
     /// <summary>
-    /// Adds extensions that allow the filling of polygon outlines to the <see cref="Image{TPixel}"/> type.
+    /// Adds extensions that allow the filling of polygon outlines.
     /// </summary>
     public static class FillPathExtensions
     {
         /// <summary>
-        /// Flood fills the image in the shape of the provided polygon with the specified brush.
+        /// Flood fills the image in the shape of the provided polygon with the specified brush..
```
tocsoft, line 15:
```suggestion
        /// Flood fills the image in the shape of the provided polygon with the specified brush.
```

src/ImageSharp.Drawing/Processing/Processors/Drawing/RecursiveImageProcessor.cs
```diff
@@ -0,0 +1,101 @@
...
+
+using System;
+using SixLabors.ImageSharp.PixelFormats;
+using SixLabors.ImageSharp.Processing;
+using SixLabors.ImageSharp.Processing.Processors;
+
+namespace SixLabors.ImageSharp.Drawing.Processing.Processors.Drawing
+{
+    /// <summary>
+    /// Allows the recursive application of processing operations against and image.
```
tocsoft, line 12:
```suggestion
    /// Allows the recursive application of processing operations against an image.
```

src/ImageSharp.Drawing/Processing/Processors/Drawing/RecursiveImageProcessor.cs
```diff
@@ -0,0 +1,101 @@
...
+                this.sourceRectangle = sourceRectangle;
+            }
+
+            public void Dispose()
+            {
+            }
+
+            public void Execute()
+            {
+                // Perform any processing logic you want here, in this case we are going to
```
tocsoft, line 79: Comment doesn't read well.... most of it can probably go as the code is relatively self explanatory.... 

I would delete the 3 lines before 
```
// so clone out our source image so we can apply
 // various effects to it without mutating the origional yet.
```

src/ImageSharp.Drawing/Processing/Processors/Drawing/RecursiveImageProcessor.cs
```diff
@@ -0,0 +1,101 @@
...
+                // down and using an image brush to draw that portion the the original image
+                // so clone out our source image so we can apply
+                // various effects to it without mutating the origional yet.
+                using Image<TPixel> clone = this.source.Clone(this.recursiveImageProcessor.Operation);
+
+                // Crop it down to just the size of the shape
+                clone.Mutate(x => x.Crop((Rectangle)this.recursiveImageProcessor.Path.Bounds));
+
+                // Use an image brush to apply cloned image as the source for filling the shape
+                var brush = new ImageBrush(clone);
```
tocsoft, line 90: We should be able to add an overload of ImageBrush that means we can avoid the crop operation and save memory... it just needs an area of interest of the source/cloned image that its allowed to point with.

i.e. `var brush = new ImageBrush(clone, (Rectangle)this.recursiveImageProcessor.Path.Bounds);`

JimBobSquarePants: I’ll have a look 👍

JimBobSquarePants: Done!

src/ImageSharp.Drawing/Processing/Processors/Drawing/RecursiveImageProcessor.cs
JimBobSquarePants: Done!

src/ImageSharp.Drawing/Processing/Processors/Drawing/RecursiveImageProcessor.cs
```diff
@@ -0,0 +1,97 @@
...
+            public void Execute()
+            {
+                // Clone out our source image so we can apply various effects to it without mutating
+                // the original yet.
+                using Image<TPixel> clone = this.source.Clone(this.recursiveImageProcessor.Operation);
+
+                // Use an image brush to apply cloned image as the source for filling the shape.
+                // We pass explicit bounds to avoid the need to crop the clone;
+                RectangleF bounds = this.recursiveImageProcessor.Path.Bounds;
+                var brush = new ImageBrush(clone, new RectangleF(0, 0, bounds.Width, bounds.Height));
```
tocsoft, line 86: 0,0 doesn't looks right to me here i would have expected it to have been the top left bound of the Path, so if the shape was Rectangle of 100x100 being applied at 40, 50, then I would have expected we would have to pass in  `(40, 50, 100, 100)`.

tocsoft: I need to try and get my head /debug the changes you made to the Brush to see what's going on but it reads here as a bug as the original code was cropping down to the truncated bounds directly 

https://github.com/SixLabors/Samples/blob/38faba5caf866c0168ea67a5e7c37c06d3043775/ImageSharp/CustomImageProcessor/Program.cs#L102

JimBobSquarePants: That’s exactly how I thought it would work but weirdly it didn’t? That’s why I added those offset tests to be sure.

JimBobSquarePants: @antonfirsov When you're back can you please sense check this. It certainly seems to work but it IS confusing. Thanks!

antonfirsov: Yeah back from vacation since yesterday, I will finally get to this PR this week.

tocsoft: ok I think I've figured out what's wrong with this.. you tried to reuse a region property that was already inside the `ImageBrushApplicator` however that region is specifically supposed to be the target region we well be drawing to not the region of the source image we are cloning from... it happened to work out for you when you did it ion this case due to the fact its ultimately the same source image with and without effects being applied, but ultimately its mixing 2 concerns and it won't work as expected for usages where the image bush is using a source image smaller than the final target where repeating texture output is supposed to be output. I'm looking at exactly what changes are needed inside the brush to make this work correctly and ill create a PR targeting this branch with the required changes.

tocsoft: #154 has the correct fix

JimBobSquarePants: Perfect! Thanks!

src/ImageSharp.Drawing/Processing/Extensions/ClearPathExtensions.cs
```diff
@@ -1,79 +1,84 @@
...
-            return new DrawingOptions(options, shapeOptions.ShapeOptions, shapeOptions.TextOptions, shapeOptions.Transform);
-        }
+        /// <param name="source">The image processing context.</param>
+        /// <param name="color">The color.</param>
+        /// <param name="path">The logic path.</param>
+        /// <returns>The <see cref="IImageProcessingContext"/> to allow chaining of operations.</returns>
+        public static IImageProcessingContext Clear(
+            this IImageProcessingContext source,
+            Color color,
+            IPath path) =>
```
antonfirsov, line 23: I would consider renaming this to `polygon`.

src/ImageSharp.Drawing/Processing/Extensions/ClearPathExtensions.cs
```diff
@@ -1,79 +1,84 @@
...
 {
     /// <summary>
-    /// Adds extensions that allow the filling of polygon outlines to the <see cref="Image{TPixel}"/> type.
+    /// Adds extensions that allow the flood filling of polygon outlines without blending.
     /// </summary>
     public static class ClearPathExtensions
     {
         /// <summary>
-        /// Clones the shape graphic options and applies changes required to force clearing.
+        /// Flood fills the image in the path of the provided polygon with the specified brush without any blending.
```
antonfirsov, line 14:
```suggestion
        /// Flood fills the image within the provided region defined by an <see cref="IPath"/> using the specified brush without any blending.
```

src/ImageSharp.Drawing/Processing/Extensions/ClearPathExtensions.cs
antonfirsov, line 23: I would consider renaming these parameters to `region` so the name better represents the role of the parameter. Also, we definitely need more descriptive docs than "The logic path.".

```suggestion
        /// <param name="region">The <see cref="IPath"/> defining the region to fill.</param>
        /// <returns>The <see cref="IImageProcessingContext"/> to allow chaining of operations.</returns>
        public static IImageProcessingContext Clear(
            this IImageProcessingContext source,
            Color color,
            IPath region) =>
```

src/ImageSharp.Drawing/Processing/Processors/Drawing/FillPathProcessor.cs
```diff
@@ -32,9 +36,9 @@ public FillPathProcessor(DrawingOptions options, IBrush brush, IPath shape)
         public IBrush Brush { get; }
 
         /// <summary>
-        /// Gets the region that this processor applies to.
+        /// Gets the logic path that this processor applies to.
```
antonfirsov, line 39:
```suggestion
        /// Gets the <see cref="Path"/> defining the region to fill.
```

src/ImageSharp.Drawing/Processing/Processors/Drawing/FillPathProcessor.cs
```diff
@@ -32,9 +36,9 @@ public FillPathProcessor(DrawingOptions options, IBrush brush, IPath shape)
...
         /// </summary>
-        public IPath Shape { get; }
+        public IPath Path { get; }
```
antonfirsov, line 41: I think "region" is a better name when it's about filling:

```suggestion
        public IPath Region { get; }
```

src/ImageSharp.Drawing/Processing/Processors/Drawing/RecursiveImageProcessor.cs
```diff
@@ -0,0 +1,97 @@
...
+using SixLabors.ImageSharp.PixelFormats;
+using SixLabors.ImageSharp.Processing;
+using SixLabors.ImageSharp.Processing.Processors;
+
+namespace SixLabors.ImageSharp.Drawing.Processing.Processors.Drawing
+{
+    /// <summary>
+    /// Allows the recursive application of processing operations against an image.
+    /// </summary>
+    public class RecursiveImageProcessor : IImageProcessor
```
antonfirsov, line 14: `ProcessInsideRegionProcessor`?

src/ImageSharp.Drawing/Processing/Extensions/ClipPathExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+    public static class ClipPathExtensions
+    {
+        /// <summary>
+        /// Applies the processing operation against a given clipping path.
+        /// </summary>
+        /// <param name="source">The image processing context.</param>
+        /// <param name="path">The target path to operate within.</param>
+        /// <param name="operation">The operation to perform on the source.</param>
+        /// <returns>The <see cref="IImageProcessingContext"/> to allow chaining of operations.</returns>
+        public static IImageProcessingContext Clip(
```
antonfirsov, line 22: I think `Clip` is a very generic overloaded and ambiguous name, doesn't help API discoverability in this case.

I wouldn'be afraid of going verbose here, for example `ProcessInsideRegion` would be a better name for this.

src/ImageSharp.Drawing/Processing/Processors/Drawing/RecursiveImageProcessor.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        /// <summary>
+        /// The main workhorse class. This has access to the pixel buffer but
+        /// in an abstract/generic way.
+        /// </summary>
+        /// <typeparam name="TPixel">The type of pixel.</typeparam>
+        private class RecursiveImageProcessorInner<TPixel> : IImageProcessor<TPixel>
+            where TPixel : unmanaged, IPixel<TPixel>
+        {
+            private readonly RecursiveImageProcessor recursiveImageProcessor;
```
antonfirsov, line 60: We are using the term `definition` or `processorDefinition` on other places. Describes the role between the non-generic and the generic thing better IMO.

src/ImageSharp.Drawing/Processing/Processors/Drawing/RecursiveImageProcessor.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        /// <summary>
+        /// Gets the drawing options.
+        /// </summary>
+        public DrawingOptions Options { get; }
+
+        /// <summary>
+        /// Gets the logic path.
+        /// </summary>
+        public IPath Path { get; }
```
antonfirsov, line 37:
```suggestion
        public IPath Region { get; }
```

antonfirsov (commented):
Implementation and tests look good.

We need to clean up the terminology used in API naming and docs, because currently it feels confusing to me. The problem is that `IPath` can be used for many things in the Drawing library, it's name doesn't carry the purpose (which I personally dislike, but it's a common practice in drawing libraries). My recommendation is to use the term "Region" (or an equivalent unambiguously descriptive term) for an `IPath` defining a region to fill, or boundary for a recursive processor.

We also need to find a better name for the new `Clip` method.

JimBobSquarePants:
@antonfirsov I've applied a lot of renaming and updated XML docs based on your suggestions. I found that I was straying into #94 territory though as I waded through each connected method so have stopped where I am for now. 

I maintained `Clip` as the method name (_I think people will get it_) and renamed the processor `ClipPathProcessor`

JimBobSquarePants:
I'm merging this. We can do a final API review before RC1

PR #154, opened 2021-07-23
enable image brush to be only draw a portion of the source image

tocsoft:
@JimBobSquarePants  This is just the changes for you #150 branch that correctly clips the source image while still supporting target/destination offsetting (that's what the pre-existing region was for).

NOTE@ This is targeting branch of #150


JimBobSquarePants:
Ah excellent! We’ll done figuring this out!

PR #157, opened 2021-08-02
Update samples and add to main solution

JimBobSquarePants:
### Prerequisites

### Description

Updates the samples project to use the latest API and includes it in the solution so we don't forget about it in the future.

### DIFF
--- a/samples/DrawShapesWithImageSharp/ImageSharpLogo.cs
+++ b/samples/DrawShapesWithImageSharp/ImageSharpLogo.cs
@@ -1,7 +1,12 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
 using System;
 using System.Collections.Generic;
 using System.Numerics;
 using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.Drawing;
+using SixLabors.ImageSharp.Drawing.Processing;
 using SixLabors.ImageSharp.PixelFormats;
 using SixLabors.ImageSharp.Processing;
 
@@ -16,13 +21,15 @@ namespace SixLabors.Shapes.DrawShapesWithImageSharp
 
             var center = new Vector2(603);
 
-            // segment whose center of rotation should be 
+            // segment whose center of rotation should be
             var segmentOffset = new Vector2(301.16968f, 301.16974f);
-            IPath segment = new Polygon(new LinearLineSegment(new Vector2(230.54f, 361.0261f), new System.Numerics.Vector2(5.8641942f, 361.46031f)),
-                new CubicBezierLineSegment(new Vector2(5.8641942f, 361.46031f),
-                new Vector2(-11.715693f, 259.54052f),
-                new Vector2(24.441609f, 158.17478f),
-                new Vector2(78.26f, 97.0461f))).Translate(center - segmentOffset);
+            IPath segment = new Polygon(
+                new LinearLineSegment(new Vector2(230.54f, 361.0261f), new Vector2(5.8641942f, 361.46031f)),
+                new CubicBezierLineSegment(
+                    new Vector2(5.8641942f, 361.46031f),
+                    new Vector2(-11.715693f, 259.54052f),
+                    new Vector2(24.441609f, 158.17478f),
+                    new Vector2(78.26f, 97.0461f))).Translate(center - segmentOffset);
 
             // we need to create 6 of theses all rotated about the center point
             var segments = new List<IPath>();
@@ -33,13 +40,14 @@ namespace SixLabors.Shapes.DrawShapesWithImageSharp
                 segments.Add(s);
             }
 
-            var colors = new List<Rgba32>() {
-                Rgba32.FromHex("35a849"),
-                Rgba32.FromHex("fcee21"),
-                Rgba32.FromHex("ed7124"),
-                Rgba32.FromHex("cb202d"),
-                Rgba32.FromHex("5f2c83"),
-                Rgba32.FromHex("085ba7"),
+            var colors = new List<Rgba32>()
+            {
+                Color.ParseHex("35a849"),
+                Color.ParseHex("fcee21"),
+                Color.ParseHex("ed7124"),
+                Color.ParseHex("cb202d"),
+                Color.ParseHex("5f2c83"),
+                Color.ParseHex("085ba7"),
             };
 
             var scaler = Matrix3x2.CreateScale(scalingFactor, Vector2.Zero);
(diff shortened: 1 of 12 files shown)


samples/DrawShapesWithImageSharp/Program.cs
```diff
@@ -46,111 +52,107 @@ private static void OutputStars()
...
         {
-            FontFamily fam = SixLabors.Fonts.SystemFonts.Find("Arial");
-            Font font = new Font(fam, 30);
-            RendererOptions style = new RendererOptions(font, 72)
+            FontFamily fam = SystemFonts.Find("Arial");
+            var font = new Font(fam, 30);
+            var style = new RendererOptions(font, 72)
             {
-                WrappingWidth = path.Length,
+                WrappingWidth = path.Bounds.Width,
```
tocsoft, line 80: this isn't the same thing.... (if i'm reading this right this is drawing along a path)

Imagine this path was a circle, the length of the circle in this case would be its circumference, whereas the width would just be its diameter meaning the text would only get partly around the outside before wrapping and not once it closes in itself.

tocsoft (commented):
generally fine but looks like you have issue with the draw along path sample... looks like you have incorrectly changed the semantics.

samples/DrawShapesWithImageSharp/Program.cs
JimBobSquarePants: Length isn’t a property anymore so how would I calculate it?

antonfirsov: An approximation is a sum of [distances](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector2.distance?view=net-5.0) between `path.Flatten().Single().Points`. (I assume Flatten returns a single `ISimplePath` for this, but I may be wrong.)

tocsoft: without the limitation on single path... 
```c#
static float GetLength(IPath path)
{
    float dist = 0;
    foreach (var s in path.Flatten())
    {
        var points = s.Points.Span;
        if (points.Length < 2)
        {
            // only a point
            continue;
        }

        for (var i = 1; i < points.Length; i++)
        {
            dist += Vector2.Distance(points[i - 1], points[i]);
        }

        if (s.IsClosed)
        {
            dist += Vector2.Distance(points[0], points[points.Length - 1]);
        }
    }

    return dist;
}
```
we should/could probably add this this as an extension on `IPath`, or re add it back to the `IPath` interface.

definitely makes the sample you threw into slack a little less broken.

antonfirsov: Unless we had the implementations calculating the length analytically, this should be an extension method. I would name it `.CalculateApproximateLength()`.

JimBobSquarePants: As I recall there were issues calculating it as a property in ComplexPolygon. I’d rather avoid re-adding it to the interface and use an extension.

samples/DrawShapesWithImageSharp/ImageSharpLogo.cs
```diff
@@ -33,23 +40,24 @@ public static void SaveLogo(float size, string path)
...
             }
 
-            var colors = new List<Rgba32>() {
-                Rgba32.FromHex("35a849"),
-                Rgba32.FromHex("fcee21"),
-                Rgba32.FromHex("ed7124"),
-                Rgba32.FromHex("cb202d"),
-                Rgba32.FromHex("5f2c83"),
-                Rgba32.FromHex("085ba7"),
+            var colors = new List<Rgba32>()
```
antonfirsov, line 43: I think this should work:

```suggestion
            var colors = new List<Color>()
```

JimBobSquarePants: Ah yeah, forgot that.

samples/DrawShapesWithImageSharp/Program.cs
JimBobSquarePants: I went with `ComputeLength()` to match the DirectX method that does the same thing.

https://docs.microsoft.com/en-us/windows/win32/direct2d/id2d1geometry-computelength

tests/ImageSharp.Drawing.Tests/Drawing/Paths/ComputeLength.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the Apache License, Version 2.0.
+
+using Xunit;
+
+namespace SixLabors.ImageSharp.Drawing.Tests.Drawing.Paths
+{
+    public class ComputeLength
+    {
+        [Fact]
+        public void CanComputeUnrolledLength()
```
tocsoft, line 11: We probably want a test that verifies the length calculates as expected for more complex paths too... i.e. any `IPath` where the `Flatten()` call returns multiple `ISimplePath`s

JimBobSquarePants: I looked for and couldn’t find one in the codebase from before `IPath.Length` was removed to reuse. With nothing to verify against I’m just been guessing the result.

JimBobSquarePants: I'm going to need a little help here. 

I used LottieSharp to calculate a polygon length since it gave me access to the original DX method and attempted to replicate a polygon to match another test one. 

```c#
var p = new LottieSharp.Path();
p.MoveTo(15, 0);
p.LineTo(25, 0);
p.MoveTo(15, 30);
p.LineTo(25, 0);
p.Close();

using (var factory = new SharpDX.Direct2D1.Factory(SharpDX.Direct2D1.FactoryType.SingleThreaded))
{
	p.GetGeometry(factory).ComputeLength().Dump(); // 73.24555
}

```
Our test code.

```c#
[Fact]
public void CanComputeUnrolledLength()
{
    var pb = new PathBuilder();
    pb.AddLine(new Vector2(15, 0), new Vector2(25, 0))
        .AddLine(new Vector2(15, 30), new Vector2(25, 30))
        .CloseFigure();

    IPath p = pb.Build();
    Assert.Equal(73.24555, p.ComputeLength()); // 83.24555206298828
}
```
We seem to be calculating a value exactly 10 more than the DX. I couldn't figure out a way to convert their types to ours to verify the output.

Tbh though, I don't understand how either is getting that number. The path should be a rectangle of sides 10x30 which would make the length 80 shouldn't it?

SkiaSharp seems to give me junk. Not sure what I'm doing wrong there. 

```c#
using var p = new SKPath();
p.MoveTo(15, 0);
p.LineTo(25, 0);
p.MoveTo(15, 30);
p.LineTo(25, 0);
p.Close();

var measure = new SKPathMeasure(p, false, 1);
measure.Length.Dump(); // 10

```

tocsoft: I feel your overthinking this too much..... all you want is:

```c#
[Fact]
public void CanComputeUnrolledLengthComplexPath()
{
    var polygon = new ComplexPolygon(
        new RectangularPolygon(PointF.Empty, new PointF(100, 200)),
        new RectangularPolygon(new PointF(1000, 1000), new PointF(1100, 1200)));

    Assert.Equal(1200, polygon.ComputeLength());
}
```

this just ensures that `Flatten()` returns multilple `ISimplePath`s and that we sum the distances across all the `ISimplePath`s rather than just the first or any other logic.

tocsoft: btw your use of path builder isn't the same in your 2 samples. we don't have a move line equivalent option, the 2nd add line call adds a line joining the end of the first and the start of the 2nd and then the CloseFigure joins the start and the end, which explains the confusion in the length differences.

JimBobSquarePants: Oh I'm definitely overthinking things! 😄 

We should probably add  `MoveTo` and `LineTo` methods then shouldn't we (_or have I missed an equivalent_)? 

I'm still intrigued by the result. I must be visualizing something badly because it seemed like a rectangle to me.

tocsoft (approved):

PR #178, opened 2021-11-22
Fix accuracy of drawing text along a path

### LINKED ISSUE #159: Improve accuracy of drawing text along a path
Modify `PathGlyphBuilder`

https://github.com/SixLabors/ImageSharp.Drawing/blob/653cf5f8f6249c8e5432ad992e42a721d8638003/src/ImageSharp.Drawing/Shapes/Text/PathGlyphBuilder.cs#L39-L49

To more accurately transform individual glyph paths.

As demonstrated in the samples project the glyphs are not evenly aligned along the path.

https://github.com/SixLabors/ImageSharp.Drawing/blob/653cf5f8f6249c8e5432ad992e42a721d8638003/samples/DrawShapesWithImageSharp/Program.cs#L68-L81

Hello World Hello World Hello World Hello World Hello World Hello World Hello World

This article describes how to accurately draw text along a path.
http://www.planetclegg.com/projects/WarpingTextToSplines.html

JimBobSquarePants:
### Prerequisites

- [ ] I have provided test coverage for my change (where applicable)

### Description

My attempt to fix the accuracy of drawing text along a given path. Fixes #159 

- [x] Fix horizontal text. (_Rotates along bottom center and handles vertical/horizontal layout options_)
- [x] Fix vertical layout.
- [x] Create tests

Layout rules are different for vertical text. https://svgwg.org/svg2-draft/text.html#TextpathLayoutRules
> For vertical text layout flows, the default orientation for a given glyph is along the vector that starts at the intersection point on the path to which the glyph is attached and which points in the direction 180 degrees from the angle of the curve at the intersection point.

image

Merging #177 already [improved](https://github.com/SixLabors/ImageSharp.Drawing/issues/159#issuecomment-972788669) the output for horizontal layout but more work was required to fix the over rotation of letters.

**Old**
old

**New**
new-hor

I can't tell whether the glyph offsetting in the outer row is expected.


JimBobSquarePants:
OK. So here's the current layout state for both horizontal and vertical text in this PR.

I'm rendering multiline text which is wraps at the length of the path. It looks to me like our calculations start of accurate both become less so. I'll need help figuring out why.

**Horizontal**
horizontal

**Vertical**
vertical

JimBobSquarePants:
@SixLabors/core 

Question. How, from a maths perspective would you adjust the position and rotation to compensate for the vertical distance of each glyph away from the path? Ideally the letter spacing should never change.

THEQUI~1

JimBobSquarePants:
I'm actually very happy with this now however there is an issue with `TextBuilder_Bounds_AreCorrect` since we now don't skip empty glyphs in the new Fonts version. @tocsoft @antonfirsov any tips here on how to fix that?
CanDrawTextAlongPathHorizontal_Rgba32_Blank100x100_type-spiral
CanDrawTextAlongPathHorizontal_Rgba32_Blank120x120_type-triangle
CanDrawTextAlongPathHorizontal_Rgba32_Blank350x350_type-circle
CanDrawTextAlongPathVertical_Rgba32_Blank250x250_type-triangle
CanDrawTextAlongPathVertical_Rgba32_Blank350x350_type-circle

tocsoft:
To be honest I'm not exactly convinced how great a test that is in the first place as they are measuring different things at different points in the processing pipeline of rendering, one don prior to hinting and everything the other is a final vector once all that stuff is applied etc

to unblock yourself in this test case however you can do 

```diff
-FontRectangle measuredBounds = TextMeasurer.MeasureBounds(text, options);
+FontRectangle direcectMesured = TextMeasurer.MeasureBounds(text, options);
+var measuredBounds = new FontRectangle(new(0, 0), direcectMesured.Size + direcectMesured.Location);
```
and that test passes again

JimBobSquarePants:
I have no idea what is going on here with `r`. It only happens with ellipses using vertical mode and multiline text. You can see things realign with `o` go out of sync with `u` then every following glyph is correct. All glyphs are center aligned when using vertical mode as shown in the second image.

image

image

src/ImageSharp.Drawing/Shapes/Path.cs
```diff
@@ -265,18 +265,16 @@ public static bool TryParseSvgPath(ReadOnlySpan<char> svgPath, out IPath value)
...
-                        svgPath = TrimSeparator(svgPath);
-                        svgPath = FindScaler(svgPath, out float angle);
-                        svgPath = TrimSeparator(svgPath);
-                        svgPath = FindScaler(svgPath, out float largeArc);
-                        svgPath = TrimSeparator(svgPath);
-                        svgPath = FindScaler(svgPath, out float sweep);
-
-                        svgPath = FindPoint(svgPath, out PointF point, relative, c);
-                        if (svgPath.Length > 0)
+                        if (TryFindScaler(ref svgPath, out float radiiX)
```
JimBobSquarePants, line 268: Skia was comparing each pointer here whereas we would only draw an arc if the span processing wasn't complete. Wile fixing to allow checking each step I change the order of params for consistency.

src/ImageSharp.Drawing/Shapes/Text/PathGlyphBuilder.cs
```diff
@@ -30,22 +33,53 @@ public PathGlyphBuilder(IPath path)
...
-            SegmentInfo point = this.path.PointAlongPath(rect.Left);
+            // Find the intersection point.
+            // This should be offset to ensure we rotate at the bottom-center of the glyph.
+            float halfWidth = bounds.Width * .5F;
+
+            // Find the point of this intersection along the given path.
+            SegmentInfo pathPoint = this.path.PointAlongPath(bounds.Left + halfWidth);
 
-            Vector2 targetPoint = point.Point + new PointF(0, rect.Top - this.yOffset);
+            // Now offset our target point since we're aligning the bottom-left location of our glyph against the path.
```
JimBobSquarePants, line 74: The SVG docs say vertical layout should use the centre point but I found left/bottom still yielded better results.

src/ImageSharp.Drawing/ImageSharp.Drawing.csproj
```diff
@@ -17,8 +17,8 @@
     <None Include="..\..\shared-infrastructure\branding\icons\imagesharp.drawing\sixlabors.imagesharp.drawing.128.png" Pack="true" PackagePath="" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="SixLabors.Fonts" Version="1.0.0-beta16" />
-    <PackageReference Include="SixLabors.ImageSharp" Version="2.1.1-alpha.0.1" />
+    <PackageReference Include="SixLabors.Fonts" Version="1.0.0-beta16.13" />
```
JimBobSquarePants, line 20: All the text drawing reference updates are a result of updating Fonts which contains layout fixes.

tocsoft (approved):

Issue #158, opened 2021-08-02
Expose JointStyle and EndCapStyle on IPen


antonfirsov:
Modify `DrawPathProcessor`:
https://github.com/SixLabors/ImageSharp.Drawing/blob/b58a245c75d8b546750fbabd9d289c4c7ac513a4/src/ImageSharp.Drawing/Processing/Processors/Drawing/DrawPathProcessor.cs#L52

And probably `DrawTextProcessor`:
https://github.com/SixLabors/ImageSharp.Drawing/blob/b58a245c75d8b546750fbabd9d289c4c7ac513a4/src/ImageSharp.Drawing/Processing/Processors/Text/DrawTextProcessor%7BTPixel%7D.cs#L342-L354

to use the new properties.

This will simplify life for use cases like #155.

Adding to 1.0 with lowest priority for now, @JimBobSquarePants feel free to kick it out of scope if it becomes blocking.

antonfirsov:
@jtjackson thanks for your help with this!

PR #185, opened 2021-12-18
Expose JointStyle and EndCapStyle on IPen

jtjackson:
### Prerequisites

### Description

My implementation of exposing two properties on IPen. Addresses #158 
I still need to add some tests to cover the new properties exposed here.

This is my first attempt at contributing to ImageSharp so be patient with me. 🙂  

Any feedback is appreciated.


src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs
```diff
@@ -61,6 +61,19 @@ public static IPath GenerateOutline(this IPath path, float width, float[] patter
...
+        /// Generates a outline of the path with alternating on and off segments based on the pattern.
+        /// </summary>
+        /// <param name="path">the path to outline</param>
+        /// <param name="width">The final width outline</param>
+        /// <param name="pattern">The pattern made of multiples of the width.</param>
+        /// <param name="jointStyle">The style to render the joints.</param>
+        /// <param name="patternSectionCapStyle">The style to render between sections of the specified pattern.</param>
+        /// <returns>A new path representing the outline.</returns>
+        /// <exception cref="ClipperException">Couldn't calculate offset.</exception>
+        public static IPath GenerateOutline(this IPath path, float width, ReadOnlySpan<float> pattern, JointStyle jointStyle, EndCapStyle patternSectionCapStyle)
```
JimBobSquarePants, line 74: We'll need to review these extension methods to ensure that the method chaining covers all sensible combinations. I've done others in the library but not this one yet I think.

jtjackson:
My apologies.  I didn't see the comments until now. :( 

I'll follow up tonight. Thanks!

JimBobSquarePants:
>My apologies. I didn't see the comments until now. :(

No worries mate! We should all have been taking some time off anyway 😄

src/ImageSharp.Drawing/Processing/IPen.cs
```diff
@@ -24,5 +24,15 @@ public interface IPen
         /// Gets the stoke pattern.
         /// </summary>
         ReadOnlySpan<float> StrokePattern { get; }
+
+        /// <summary>
+        /// Gets the stroke joint style
+        /// </summary>
+        public JointStyle StrokeJoint { get; }
```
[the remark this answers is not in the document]
jtjackson: Yeah, usually, I keep the enum type name and property name separate, but as it seems we are only using this once, I'm fine with matching them up here. I'm not feeling creative right now. 🙂

src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs
jtjackson: That's what I've been trying to figure out.  I'm still navigating the tests that we have and determining what I need to do to add to the test coverage.  

I know that there's a `GenerateOutline` usage that you referred to in https://github.com/SixLabors/ImageSharp.Drawing/discussions/155#discussioncomment-1120271:

https://github.com/SixLabors/ImageSharp.Drawing/blob/3ce36e0ac80036c9cb60ba6c12f99cce001f71dd/samples/DrawShapesWithImageSharp/Program.cs#L196-L204

So, I'm trying to wrap my head around how we can use these new extension methods.  Should I add more to the sample program?

jtjackson:
I added some tests for the new properties. I also took the liberty of updating the sample code so it will use the updated `GenerateOutline` extension method.

The PR is ready for review.  If there is anything I may have missed or overlooked, please let me know. :-)

jtjackson:
Looks like checks failed due to code coverage. I am not familiar with CodeCov here, so can someone clarify what is needed to make this pass?

antonfirsov:
@jtjackson you need to authenticate CodeCov with github to see the [coverage change report](https://app.codecov.io/gh/SixLabors/ImageSharp.Drawing/compare/185/changes). Coverage for a `GenerateOutline` overload call chain has been lost because you are no longer calling the top level overload. It's our fault that we don't have any direct tests for `GenerateOutline`, so don't worry about it

src/ImageSharp.Drawing/Processing/Pen.cs
```diff
@@ -74,5 +74,11 @@ public Pen(IBrush brush, float width)
 
         /// <inheritdoc/>
         public ReadOnlySpan<float> StrokePattern => this.pattern;
+
+        /// <inheritdoc/>
+        public JointStyle JointStyle { get; }
+
+        /// <inheritdoc/>
+        public EndCapStyle EndCap { get; }
```
antonfirsov, line 82: I don't see any ways to change these properties.

antonfirsov (commented):
We need tests validating the end-to end capability, meaning that the test should actually draw lines with custom `JointStyle` and `EndCapStyle` and demonstrate we can solve problems like #155.

You can take a look at [DrawPathTests.cs](https://github.com/SixLabors/ImageSharp.Drawing/blob/master/tests/ImageSharp.Drawing.Tests/Drawing/DrawPathTests.cs). When running in Debug, the `RunValidatingProcessorTest`method should save debug output to `tests/Images/ActualOutput/Drawing/DrawLinesTests/`. Initially it will fail. If you are fine with the result output you can copy the images to the [matching ReferenceOutput folder](https://github.com/SixLabors/ImageSharp.Drawing/tree/master/tests/Images/ReferenceOutput/Drawing/DrawLinesTests), making them pass. You should push the reference output to GIT, but make sure the images are not crazy huge.

src/ImageSharp.Drawing/Processing/Pen.cs
jtjackson: I was following what was used for `StokeWidth` and `StrokePattern` properties.  We can set the properties via extension methods like `GenerateOutline`.

Would it be ideal to be able to update the properties in a way like this:

```
Pen pen = new Pen();

pen.JointStyle = JointStyle.Miter;
```

Something like that?

antonfirsov: We initialize `StokeWidth` and `StrokePattern` via costructor. If we want to follow existing patterns, we need to initialize it with in a new constructor overload ... which is an ugly bloat.

Though I'm tempted to change the whole thing and remove readonly nature of `Pen`, remove all the constructors, and make every property settable.

@JimBobSquarePants @tocsoft thoughts?

antonfirsov: Pro: no constructor bloat
Con: loosing API symmetry with brushes which are readonly.

antonfirsov: @jtjackson don't be blocked by our API design discussion, just pick whatever you like and go ahead with the tests, as soon as we decide we can easily refactor.

JimBobSquarePants: Yeah... I wanna do a usability review to make sure the APIs are where we need them to be. We can change things there.

antonfirsov: Opened #195. Even if it's not where we land in the end, @jtjackson I recommend to go with the settable properties for `JointStyle` and `EndCapStyle` for simplicity.

jtjackson:
Just made the new properties settable. We can change this as needed based on the discussion in #195.  

I also uploaded the reference images for the tests in [DrawLinesTests.cs](https://github.com/SixLabors/ImageSharp.Drawing/blob/master/tests/ImageSharp.Drawing.Tests/Drawing/DrawLinesTests.cs). I don't think it's too big (6 KB) but let me know if that's an issue.

It would be pretty too easy for me to go crazy and add a lot more tests (once a QA Engineer, always a QA engineer 🙂) but I think it's a good start unless we think we should add more.  Right now, we can see where the `JointStyle` and `EndCap` properties are rendered based on the reference images (it's more clear with anti-alias turned on, I would think).

src/ImageSharp.Drawing/Processing/Pen.cs
antonfirsov: No need for more tests, but I would make the lines somewhat ticker for the JointStyle + EndCap tests for tow reasons:
- Make the results more visible + auditable (currently the difference between JointStyle = Round vs Square is not visible)
- Make sure it fails when there is a bug, even if the comparison is tolerant

jtjackson: I just updated the images and tests so that we should be able to discern the difference between styles. Let me know if it stands out to you.

One thing I noticed is that some tests failed locally but I can't tell if it's the change or it's my machine.  I did try testing against master and it passed.  🤔 I'll debug some more when I get the chance.

Would it be possible for you to pull my changes and run tests on your machine? I suspect that the CI check will also let us know whether my changes passed muster.

jtjackson:
I finally figured out why some of the tests were failing. It was because I added `EndCapStyle` parameter here:
https://github.com/SixLabors/ImageSharp.Drawing/blob/ce91687fa96442c5c26d450d77179fec63960dcf/src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs#L88-L93

Which would have a conflict with the call below because the "default" parameter appears to be `EndCapStyle.Square`:
https://github.com/SixLabors/ImageSharp.Drawing/blob/ce91687fa96442c5c26d450d77179fec63960dcf/src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs#L200-L206

According to the enum, the default is actually `EndCapStyle.Butt`. Any reason why it's square instead of butt in the last `GenerateOutline` method here?

I just removed the `EndCapStyle` parameter so that it would allow the tests to pass, but I want to make sure that to understand what's happening here.  The reference images also have been updated and the comparison looks good but let me know.

JimBobSquarePants:
>I just removed the EndCapStyle parameter so that it would allow the tests to pass, but I want to make sure that to understand what's happening here. The reference images also have been updated and the comparison looks good but let me know.

@jtjackson  You've found a mistake. The default should always match the default in the enumeration. 

I've pulled down your fork and have given `OutlinePathExtensions` a go-over to fix up the method inheritance and documentation. I've also fixes various LFS issues where images were not stored as LFS pointers. 

I'll need to get push access to your fork to fix the LFS issues properly though so if you can do that please I can get this completed and merged.

tocsoft:
The default needs to be square, if my memory serves it to match the default style from System.Drawing. The enum isn't one we originally created and that would be why the defaults miss match. However we should probably just change the enum to match the method.

JimBobSquarePants:
Skia uses the same order as we do though.
https://docs.microsoft.com/en-us/dotnet/api/skiasharp.skstrokecap?view=skiasharp-2.80.2

System.Drawing defaults to Flat by the look of things.
https://docs.microsoft.com/en-us/dotnet/api/system.drawing.drawing2d.linecap?view=dotnet-plat-ext-6.0

jtjackson:
> I'll need to get push access to your fork to fix the LFS issues properly though so if you can do that please I can get this completed and merged.

@JimBobSquarePants I'm still learning more about how to do PRs on Github (I've done it on private repos, but not public) but it seems that if I checked `Allow edit by maintainers`, that should give you a push access to my fork? 

At any rate, I've added you as a collaborator to my forked repo. Let me know if you have any issues. 🙂

JimBobSquarePants:
@jtjackson You did everything just fine, it's just there's known [issues](https://github.com/git-lfs/git-lfs/issues/2291) with Git LFS and GitHub that can in some scenarios be worked around but adding new LFS pointers seems not to be one of those scenarios. Only full repo access seems to provide what I need to push the new pointers. 

Anyway, I'm very happy with these changes and the new tests look really good also. Thanks very much for your help 👍

JimBobSquarePants (approved):

PR #206, opened 2022-03-01
Add drawing centric path building api.

tocsoft:
Add drawing centric path building api and migrate GlyphRenderer to use new helper class directly

### Prerequisites

### Description
As we replatformed glyph rendering onto of this API its should be covered by font tests.


tocsoft:
we probably have some missing api/cases as those are just the ones we can utilise for GlyphRendering but we can expand as required.

That's the class I would hand the SVG path parsing logic off too (at least as an extension method).

src/ImageSharp.Drawing/Shapes/PathDrawer.cs
```diff
@@ -0,0 +1,106 @@
...
+// Licensed under the Apache License, Version 2.0.
+
+using System.Numerics;
+
+namespace SixLabors.ImageSharp.Drawing
+{
+    /// <summary>
+    /// Allow you to derivatively draw shapes and paths.
+    /// </summary>
+    public class PathDrawer
```
JimBobSquarePants, line 11: Is there anyway we can simplify things by combining these two types? It seems, to me, that the overlap is confusing.

tocsoft: Initially I had assumed it wouldn't work as a single class but it seems I was wrong, I've merged the `***To()` APIs into the `PathBuilder` itself.

tocsoft:
Currently missing the `ArcTo` function (but i'll add it soon) it should then have all the required features for svg path parsing which i have a 2nd branch where i'm adding that.

src/ImageSharp.Drawing/Shapes/PathDrawer.cs
JimBobSquarePants: That's great news!!

src/ImageSharp.Drawing/Shapes/PathBuilder.cs
```diff
@@ -154,11 +160,31 @@ public PathBuilder AddLines(params PointF[] points)
...
         }
 
+        /// <summary>
+        /// Draws a quadratics bezier from the current point  to the <paramref name="point"/>
+        /// </summary>
+        /// <param name="secondControlPoint">The second control point.</param>
+        /// <param name="point">The point.</param>
+        /// <returns>The <see cref="PathBuilder"/></returns>
+        public PathBuilder QuadraticBezierTo(Vector2 secondControlPoint, Vector2 point)
+            => this.AddBezier(this.currentPoint, secondControlPoint, point);
```
JimBobSquarePants, line 176: I wonder whether the two `AddBezier` methods should be renamed? `AddQuadraticBezier` `AddCubicBezier`

JimBobSquarePants, line 176: I wonder whether the two `AddBezier` methods should be renamed? `AddQuadraticBezier` `AddCubicBezier`

JimBobSquarePants:
@tocsoft How should `ArcTo` work? I might have a crack at implementing it as I want to get this and the SVG parsing PR done so I can incorporate it into the testing for #178 (_I'm actually happy with the output for that PR, I just want extra comparisons against known SVG examples_)

I wanna get the improvements to path text drawing in before #211 so we can have the same rich text functionality in both scenarios.

JimBobSquarePants:
Ah.... I see you already worked out ArcTo

https://github.com/SixLabors/ImageSharp.Drawing/commit/d3ae3c03eab317c34ecc9bb96d1653db994a66ac#diff-ef4d23a8c344c2392f8ac64a5b6bca8d1a8d3fb3eca4d8fb7ed17a28b346bfbdR229

Shall I clean that up and copy it across?

tocsoft:
please do... I had forgotten I had gotten that far with it... I think the code still need validating as right. Also it was in need of optimising it looks to me like there should be some quick Vector2 wins possible in there.

JimBobSquarePants:
Cool. Will have a look tonight

JimBobSquarePants:
I've made some progress using this as an additional reference.

https://github.com/google/skia/blob/1f193df9b393d50da39570dab77a0bb5d28ec8ef/src/core/SkPathBuilder.cpp#L437-L505

SkMatrix is whack! What a horrible API!

JimBobSquarePants:
@tocsoft I've made progress!

There's something wrong with the point generation code in `EllipticalArcLineSegment`. I've hacked together a solution to the segment adding code from SVG.Sharpie which works perfectly. Need to compare and fix our original implementation but will definitely need some help.

Black on the left represents our working solution, red the solution derived from `EllipticalArcLineSegment`. Skia is on the right which matches the SVG output.

image

src/ImageSharp.Drawing/Shapes/ArcLineSegment.cs
```diff
@@ -0,0 +1,316 @@
...
+using System.Collections.Generic;
+using System.Numerics;
+using System.Runtime.CompilerServices;
+
+namespace SixLabors.ImageSharp.Drawing
+{
+    /// <summary>
+    /// Represents a line segment that contains radii and angles that will be rendered as a elliptical arc.
+    /// </summary>
+    public class ArcLineSegment : ILineSegment
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah, meant to remove it!

JimBobSquarePants (approved):

PR #240, opened 2022-07-22
Determine minX/minY/maxX/maxY by iterating over points/paths only onc…

remcoros:
…e instead of using Linq

### Prerequisites

### Description

While investigating another (solved) issue, I saw this pattern of using Linq to get the minX/minY/length/etc. This iterates over the list multiple times and could be optimized by iterating only once and keeping track of min/max values.


JimBobSquarePants (approved):
Looks great. Thanks!

Issue #195, opened 2022-01-18
Revisit Pen API


antonfirsov:
### Constructor bloat

Currently `Pen`s are immutable. As a consequence, all properties are initialized via constructor, which leads to terrible constructor bloat, especially after adding `JointStyle` and `EndCapStyle` to `Pen` in #185.

Default arguments are also off the table because of extensibility issues and non-constant parameters.

My suggestion is to drop immutability of `Pen` together with all it's constructors.

```C#
public class Pen
{
    IBrush StrokeFill { get; set; } // defaults to black solid brush
    float StrokeWidth { get; set; }

    // Can be also ReadOnlyMemory<float> but not many users know it's assignable to float[]
    // Con: the factory methods in `Pens` will allocate an extra array to guarantee immutability of shared constants.
    float[] StrokePattern { get; set; }

    public JointStyle JointStyle { get; set; }
    public EndCapStyle EndCap { get; set; }
}
```

An alternative for the above is to introduce a `PenBuilder` class, but that feels like overcomplication of a simple problem.

### `IPen`
I also propose removing the `IPen` base interface. I don't see the value for such an abstraction, since a pen is really just a tuple of a few simple properties, there is no logic behind.

```diff
- public interface IPen { }
```

JimBobSquarePants:
Yeah. Agree here, the constructor bloat is an issue and I don't think we should create some sort of options struct toe workaround.

I think we should also seal the type.

antonfirsov:
Actually, I haven't considered an options struct, we should explore that too. It could help avoiding the Pen <-> Brush mutability asymmetry, which I don't like much, just proposed as a lesser evil compared to constructor bloat.

jtjackson:
Doesn't make sense to me to have that many constructors for 'Pen' here. Can we just go with single one like:

```cs
public class Pen
{
    public Pen(PenOptions options) { }

    IBrush StrokeFill { get; }
    float StrokeWidth { get; }
    Span<float> StrokePattern { get; }
    public JointStyle JointStyle { get; }
    public EndCapStyle EndCap { get; }
}
```

However, the first proposed approach (with no constructors) may be a better approach.

antonfirsov:
@tocsoft since you are active these days: any thoughts on this? Any alternative ideas to avoid the constructor bloat?

dlemstra:
[...]

PR #211, opened 2022-03-04
Rich text rendering

tocsoft:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp.Drawing/pulls) open
- [ ] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [ ] I have provided test coverage for my change (where applicable)

### Description

This build on the SixLabors/Fonts#251 code to extend the capability to also altering the color / outlines.
DrawRichTextWithMixOfPensAndBrushes_Solid500x200_(255,255,255,255)_RichText-F(40)

### TODO
- [x] Figure out what's causing the w/v shifting.
- [x] Change back to Fonts package once SixLabors/Fonts#251 is merged.
- [x] Ensure we have enough test coverage

UPDATE
I've (@JimBobSquarePants ) ran with this and completed the PR replacing the existing `CachedGlypRender` with a new `RichTextGlyphRender` in the process removing some old hacks that were negatively affecting the output. Output is more accurate and crisp now. This is obvious for lower font sizes.

Here's a comparison (new right, note the improved rendering on several glyphs (i, j, h, e and others) )
image

Rich text along a path is also supported

Some example output.

Rich text following a path with red strikethrough.
CanDrawRichTextAlongPathHorizontal_Solid100x100_(0,0,0,255)_RichText-Path-(spiral)

Rich text using a different pen to underline each glyph to draw a rainbow.
DrawRichTextRainbow_Solid500x300_(0,0,0,255)_RichText-Rainbow-F(40)

PR #257, opened 2023-03-04
Revisit pen api

stefannikolei:
### Prerequisites

### Description
First spike for Revisit Pen API #195. As a discussion base.

Is this how it was described in the issue or should it be totally different?


JimBobSquarePants:
I want to have a good comparison with the changes to pens in #211 There's likely some overlap.

src/ImageSharp.Drawing/Processing/Brush.cs
```diff
@@ -12,7 +12,7 @@ namespace SixLabors.ImageSharp.Drawing.Processing
     /// A brush is a simple class that will return an <see cref="BrushApplicator{TPixel}" /> that will perform the
     /// logic for retrieving pixel values for specific locations.
     /// </remarks>
-    public interface IBrush
+    public abstract class Brush
```
JimBobSquarePants, line 15: I get that we're doing this for consistency but an abstract class with no implementation code is essentially an interface.

src/ImageSharp.Drawing/Processing/PenOptions.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the Apache License, Version 2.0.
+
+using System.Collections.Generic;
+
+namespace SixLabors.ImageSharp.Drawing.Processing
+{
+    /// <summary>
+    /// Options for the Pen
+    /// </summary>
+    public class PenOptions
```
JimBobSquarePants, line 11: I wonder if we'd be better with `get; init;` patter here. I'd make `JointStyle` and `EndCapStyle` non-nullable also providing defaults.

src/ImageSharp.Drawing/Processing/Pen.cs
```diff
@@ -16,43 +16,22 @@ namespace SixLabors.ImageSharp.Drawing.Processing
...
 
-        /// <summary>
-        /// Initializes a new instance of the <see cref="Pen"/> class.
-        /// </summary>
-        /// <param name="color">The color.</param>
-        /// <param name="width">The width.</param>
-        public Pen(Color color, float width)
-            : this(new SolidBrush(color), width)
+        public Pen(PenOptions options)
+            : this(options.StrokeFill, options.StrokeWidth)
```
JimBobSquarePants, line 34: `PenOptions` contains the joint/endcap properties but we're not using them

src/ImageSharp.Drawing/Processing/PenOptions.cs
stefannikolei: Yea. But I would really go for the required keyword. But for that we need to use c#11 and I du not know if this is bound to net7

JimBobSquarePants: That's off the table unfortunately. Some stuff works, some stuff doesn't but looks like it would. I'm not sure we can take the risk.

antonfirsov: > I wonder if we'd be better with `get; init;` patter here.

init-only setters officially require .NET 5+, so we would need to drop the old targets, which should be a bigger strategic decision. I recommend to avoid doing it in this PR, and revisit the problem when we switch to .NET 6.0+.

JimBobSquarePants: Ah ok. We can do that then. Thanks!

JimBobSquarePants (approved):
OK. This looks great to me now. 👍 

Thanks for your help!

antonfirsov:
@stefannikolei any reason you made `PenOptions` a class instead of a struct?

stefannikolei:
> @stefannikolei any reason you made `PenOptions` a class instead of a struct?

Because of https://github.com/SixLabors/ImageSharp.Drawing/issues/195#issuecomment-1035991253

JimBobSquarePants:
There's a chance those options might be removed as part of #211 I'm just merging the conflicts there now. 

If not, I'll make them a struct

stefannikolei:
> > Because of https://github.com/SixLabors/ImageSharp.Drawing/issues/195#issuecomment-1035991253
> 
> 
> 
> That comment actually proposes `struct PenOptions`, or am I missing something? If we keep it after the revisit, it's better to save a heap allocation.

🙈 yeah you are absolute right.

JimBobSquarePants:
I've got it covered

PR #261, opened 2023-03-20
Upgrade to clipper 2

JimBobSquarePants:
### Prerequisites

### Description

Replaces Clipper 1 with Clipper 2 which benchmarks around 2X faster. Everythign was copied as-is with the only changes to make all the clipper code internal.


JimBobSquarePants:
We should remove Clipper from code coverage

tocsoft (dismissed):
looks good to me 👍

tocsoft (approved):

PR #263, opened 2023-03-22
Fix rendering along open paths.

JimBobSquarePants:
### Prerequisites

- [x] I have written a descriptive pull-request title
- [x] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp.Drawing/pulls) open
- [x] I have verified that I am following matches the existing coding patterns and practice as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [x] I have provided test coverage for my change (where applicable)

### Description

Fixes #234

The issue is caused by our code which calculates the position of a point along the given path at a given distance. The code works well for closed paths since it reduces the distance to the remainder (`distance %= path.Length`) which makes a nice loop.

For open paths we don't want to do that, we want to create a new virtual point at the correct angle and distance calculated from the last two points in the path.

I've also renamed `DrawLines` to `DrawLine` as you can only draw one line at a time using the method.

PR #265, opened 2023-03-30
Inline Clipper

### LINKED ISSUE #10: Add possibility to make intersection of shapes

### LINKED ISSUE #250: Flattening the result of a Clip gives different results for horizontal and vertical lines
### Description
Flattening the result of a Clip gives different results for horizontal and vertical lines.

### Steps to Reproduce
```csharp
        private static void Problem()
        {
            var clip = new PathCollection(new RectangularPolygon(new PointF(24, 16), new PointF(777, 385)));

            var vert = new SixLabors.ImageSharp.Drawing.Path(new LinearLineSegment(new PointF(26, 384), new PointF(26, 163)));
            var horiz = new SixLabors.ImageSharp.Drawing.Path(new LinearLineSegment(new PointF(26, 163), new PointF(176, 163)));

            IPath reverse =  vert.Clip(clip);
            var result1 = vert.Clip(reverse).Flatten().Select(x => x.Points);

            reverse = horiz.Clip(clip);
            var result2 = horiz.Clip(reverse).Flatten().Select(x => x.Points);

            var same = result1.Count() == result2.Count();
        }
```
I expect the Count of both result1 and result2 to be identical.

### System Configuration

- ImageSharp.Drawing version: 1.0.0-beta15
- Other ImageSharp packages and versions: ImageSharp v.2.1.3
- Environment (Operating system, version and so on): Win11 v.22H2
- .NET Framework version: net6.0

JimBobSquarePants:
### Prerequisites

### Description

- Introduces a port of the bit of Clipper2 we use that uses Vector2 throughout. This uses half the memory of the default implementation, brings a minor (_~5ms for large geo polygon test_) speed improvement, plus makes the code much easier to maintain and investigate for optimization opportunities. 
- Exposes a `ClippingOperation` enum to `IPath.Clip` via `ShapeOptions`. Fixes #10 
- Adds tests for vertical and horizontal path clipping. Fixes #250 

#263 should be merged before this as I'll likely have to update the reference file again.

TODO;

- [x] Finish cleaning up `PolygonClipper` and `PolygonOffsetter` to match standards.
- [x] Add tests for the various `ClippingOperation` values.
- [x] Fix rounding inaccuracy on .NET 472
- [ ] Ask @antonfirsov whether `PolygonScanner` needs to support the [Clipper fill rules](http://www.angusj.com/clipper2/Docs/Units/Clipper/Types/FillRule.htm) `Positive` and `Negative`


src/ImageSharp.Drawing/Shapes/ClippingOperation.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// </summary>
+    /// <remarks>
+    /// All clipping operations except for Difference are commutative.
+    /// </remarks>
+    public enum ClippingOperation
+    {
+        /// <summary>
+        /// No clipping is performed.
+        /// </summary>
+        None,
```
JimBobSquarePants, line 17: `None` only seems useful if we allow choosing the operation via `Outline`. I don't know if it's wise to do so since Clipper uses `Difference` by default.

src/ImageSharp.Drawing/Shapes/PolygonClipper/BoundsF.cs
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System;
+using System.Numerics;
+
+namespace SixLabors.ImageSharp.Drawing.Shapes.PolygonClipper
+{
+    internal struct BoundsF
```
JimBobSquarePants, line 9: There's a good chance this can be removed and we can use `RectangleF`

JimBobSquarePants:
> > Finish cleaning up PolygonClipper and PolygonOffsetter to match standards.
> 
> I disagree this kind of refactoring is needed. We don't want to maintain (=understand + test) that code. Much better to include the original code as-is (only visibilities changed), and provide a facade on the top, so it is a no-brainer to keep it in sync with the upstream. Refactoring on the other hand gives extra work and has a non-zero chance to introduce bugs. I don't see the value, only downsides.

That horse has bolted I’m afraid. The refactor to half the memory usage and use Vector2 was significant. I’m happy to track upstream and implement any fixes.

antonfirsov:
> refactor to half the memory usage

I didn't know (or missed) that part. If so, it all makes sense. Wasn't able to look at this during the weekend, hope this week I can do it.

JimBobSquarePants:
> > refactor to half the memory usage
> 
> I didn't know (or missed) that part. If so, it all makes sense. Wasn't able to look at this during the weekend, hope this week I can do it.

No worries. I found a few things to improve upon this evening anyway.

tocsoft (approved):
lgtm 👍

PR #271, opened 2023-05-04
Support Vertical Mixed Text

JimBobSquarePants:
### Prerequisites

### Description

Updates the glyph renderers to handle support for [Unicode Standard Annex #50 Unicode Vertical Text Layout](https://www.unicode.org/reports/tr50/) 

Note: I've discovered an issue when rendering the decorations over certain CFF fonts in that holes are left, however I'm unable to determine what the cause could be. The paths themselves look ok. 

This only affects filling a glyph collection.


src/ImageSharp.Drawing/Shapes/Text/BaseGlyphBuilder.cs
```diff
@@ -147,30 +147,32 @@ public virtual void SetDecoration(TextDecorations textDecorations, Vector2 start
...
+            Vector2 c = end + pad;
+            Vector2 d = end - pad;
 
             // Drawing is always centered around the point so we need to offset by half.
             Vector2 offset = Vector2.Zero;
             if (textDecorations == TextDecorations.Overline)
             {
                 // CSS overline is drawn above the position, so we need to move it up.
-                offset = new(0, -(thickness * .5F));
+                offset = rotated ? new(thickness * .5F, 0) : new(0, -(thickness * .5F));
```
tocsoft, line 163: do we still need these offsets?, now we seem to also be applying them above? 

Reason I ask is looking at the onion skin of the changed test images shows the decorations have shifted and are less sharp (i.e. they look like they are .5 of a pixel off and thus aliased edges are happening)

tocsoft: fair enough...just figured i'd check. I'm sure we can circle back round to the minor visual differences, and as you say try and tackle that at the rasterizer level, which will improve things acros the board not just for test rendering.

tocsoft (approved):

JimBobSquarePants:
@tocsoft I couldn't help myself and implemented the smart pixel clamping I was on about in a manner that allowed preserving defined pen thickness. The results look great 😁

tocsoft (approved):
looks way better all around 👍

PR #288, opened 2023-08-16
Remove compiler conditions

JimBobSquarePants:
### Prerequisites

### Description

Removes all outdated `#if...#else` conditions from the code.


src/ImageSharp.Drawing/Shapes/PolygonClipper/PolygonOffsetter.cs
```diff
@@ -77,7 +77,7 @@ public void Execute(float delta, PathsF solution)
         this.ExecuteInternal(delta);
         if (this.groupList.Count == 0)
         {
-            goto Error;
+            return;
```
JimBobSquarePants, line 80: I spotted this when trying to run a benchmark. I'd deviated from clippers default behavior for some reason.

src/ImageSharp.Drawing/Utilities/SortUtility.KeyValueSort.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
 
+using System.Numerics;
```
JimBobSquarePants, line 4:
```
@antonfirsov I spotted this comment. Are we able to replace this code with built in APIs now?

// Adapted from:
// https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/ArraySortHelper.cs
// If targeting .NET 5, we can call span based sort, but probably not worth it only for that API.
```

dlemstra (dismissed):

src/ImageSharp.Drawing/Utilities/SortUtility.KeyValueSort.cs
antonfirsov: Yes, we can now delete `SortUtility` and use `span.Sort()` instead.

JimBobSquarePants: Awesome! I didn't even know about the additional sort APIs.

antonfirsov (approved):

PR #289, opened 2023-08-22
Enable nullability checks

JimBobSquarePants:
### Prerequisites

### Description

Removes all `#nullable disable` compiler instructions except for PolygonClipper (_which looks like a ton of work_)
Performs minor cleanup to remove IDE warnings

### DIFF
--- a/src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
+++ b/src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
@@ -1,8 +1,6 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
 
-#nullable disable
-
 using System.Numerics;
 using SixLabors.ImageSharp.Drawing.Utilities;
 
@@ -76,7 +74,7 @@ public sealed class PathGradientBrush : Brush
     }
 
     /// <inheritdoc />
-    public override bool Equals(Brush other)
+    public override bool Equals(Brush? other)
     {
         if (other is PathGradientBrush brush)
         {
@@ -123,7 +121,7 @@ public sealed class PathGradientBrush : Brush
         return new Color(colors.Select(c => (Vector4)c).Aggregate((p1, p2) => p1 + p2) / colors.Length);
     }
 
-    private static float DistanceBetween(PointF p1, PointF p2) => ((Vector2)(p2 - p1)).Length();
+    private static float DistanceBetween(Vector2 p1, Vector2 p2) => (p2 - p1).Length();
 
     private readonly struct Intersection
     {
@@ -178,13 +176,14 @@ public sealed class PathGradientBrush : Brush
 
         public Vector4 ColorAt(PointF point) => this.ColorAt(DistanceBetween(point, this.Start));
 
-        public bool Equals(Edge other)
-            => other.Start == this.Start &&
-                other.End == this.End &&
-                other.StartColor.Equals(this.StartColor) &&
-                other.EndColor.Equals(this.EndColor);
+        public bool Equals(Edge? other)
+            => other != null &&
+               other.Start == this.Start &&
+               other.End == this.End &&
+               other.StartColor.Equals(this.StartColor) &&
+               other.EndColor.Equals(this.EndColor);
 
-        public override bool Equals(object obj) => this.Equals(obj as Edge);
+        public override bool Equals(object? obj) => this.Equals(obj as Edge);
 
         public override int GetHashCode()
             => HashCode.Combine(this.Start, this.End, this.StartColor, this.EndColor);
@@ -249,7 +248,7 @@ public sealed class PathGradientBrush : Brush
         {
             get
             {
-                var point = new Vector2(x, y);
+                Vector2 point = new(x, y);
 
                 if (point == this.center)
                 {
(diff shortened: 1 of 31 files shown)


src/ImageSharp.Drawing/Shapes/Path.cs
```diff
@@ -49,7 +48,10 @@ public Path(Path path)
     /// </summary>
     /// <param name="segments">The segments.</param>
     public Path(params ILineSegment[] segments)
-        => this.lineSegments = segments ?? throw new ArgumentNullException(nameof(segments));
+    {
+        ArgumentNullException.ThrowIfNull(segments);
+        this.lineSegments = segments ?? throw new ArgumentNullException(nameof(segments));
```
stefannikolei, line 53:
```suggestion
        this.lineSegments = segments;
```

src/ImageSharp.Drawing/Shapes/Path.cs
```diff
@@ -49,7 +48,10 @@ public Path(Path path)
     /// </summary>
     /// <param name="segments">The segments.</param>
     public Path(params ILineSegment[] segments)
-        => this.lineSegments = segments ?? throw new ArgumentNullException(nameof(segments));
+    {
+        ArgumentNullException.ThrowIfNull(segments);
```
stefannikolei, line 52: I think you can even get rid of this null check. Calling this method without a parameter results in an empty array

JimBobSquarePants: Gave it a go... Analyzer doesn't like that!

stefannikolei: Hm strange. My rider is not showing me an error on that

stefannikolei: But I have more errors on my Mac.. I can not even build it correct.. Need to checkout a clean copy..

JimBobSquarePants: A disturbance in the force!

image

tocsoft: the reason you need the null check is you can call it like this which will result in `segments` being null
``` csharp
LineSergment[] segments = null;
var path = new Path(segments)
```
where as if called as `new Path()` then an empty array will be passed for you.

stefannikolei: > A disturbance in the force!
> 
> image

hmm 
image
^^

stefannikolei: > the reason you need the null check is you can call it like this which will result in `segments` being null
> 
> ```cs
> LineSergment[] segments = null;
> var path = new Path(segments)
> ```
> 
> where as if called as `new Path()` then an empty array will be passed for you.

Yeah this can happen on many path. But this will not compile when Nullable Errors are turned on.

src/ImageSharp.Drawing/Shapes/ComplexPolygon.cs
```diff
@@ -133,7 +130,7 @@ public IPath Transform(Matrix3x2 matrix)
     /// <inheritdoc />
     public IEnumerable<ISimplePath> Flatten()
     {
-        var paths = new List<ISimplePath>();
+        List<ISimplePath> paths = new();
```
mandel-macaque, line 133: I'm surprised you don't get a warning when declaring paths in a diff scope. Does it? (I'm doing a dry review without compiling).

mandel-macaque (approved):
Looks good, some minor nits regarding !=.

src/ImageSharp.Drawing/Processing/Pen.cs
```diff
@@ -23,7 +21,7 @@ namespace SixLabors.ImageSharp.Drawing.Processing;
 /// </remarks>
 public abstract class Pen : IEquatable<Pen>
 {
-    private readonly float[] pattern;
+    private readonly float[]? pattern;
```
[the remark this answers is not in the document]
JimBobSquarePants: I'll have a dig through actually. We might be better off changing it.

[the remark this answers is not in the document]
JimBobSquarePants: Yeah, much better!

src/ImageSharp.Drawing/Shapes/ComplexPolygon.cs
JimBobSquarePants: No warnings. We have explicit rules requiring `this.` for members. I imagine the compiler understands that.

src/ImageSharp.Drawing/Processing/PathGradientBrush.cs
```diff
@@ -178,13 +176,14 @@ public Vector4 ColorAt(float distance)
 
         public Vector4 ColorAt(PointF point) => this.ColorAt(DistanceBetween(point, this.Start));
 
-        public bool Equals(Edge other)
-            => other.Start == this.Start &&
-                other.End == this.End &&
-                other.StartColor.Equals(this.StartColor) &&
-                other.EndColor.Equals(this.EndColor);
+        public bool Equals(Edge? other)
+            => other != null &&
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah, I've considered this but will only enforce once I update the config. It's shared across all the projects via a Git submodule.

JimBobSquarePants:
Thanks for the reviews! Much appreciated! 🥰

PR #324, opened 2024-03-08
Mitigate issue where outline thicknesses less than 0.5 fail out generate solid shapes not outlines

### LINKED ISSUE #323: DrawPolygon with a thickness of less that 0.5 fills produces a filled polygon instead of a outline
```c#
  using var image = new Image<Rgba32>(300, 300, Color.White);
  image.Mutate(x => x.DrawPolygon(Color.RebeccaPurple, 0.4f, new PointF[] {
      new(5, 5),
      new(5, 150),
      new(150, 150),
  }));

  image.Save("path.png");
```

issue kicks in on thicknesses less than `0.5f`.

## Expected output
image
## Actual output
image

### Discussed in https://github.com/SixLabors/ImageSharp.Drawing/discussions/322

<div type='discussions-op-text'>

<sup>Originally posted by **woutware** March  8, 2024</sup>
Edit: after further investigations this seems to be a bug, I've edited the text below.

When argument thickness is less than 0.5 (e.g. 0.4), method DrawPolygon now fills the polygon, while it used to only stroke the outline, how to get the old behavior? Tested on ImageSharp 3.1.3, ImageSharp.Drawing 2.1.2. When the thickness is equal to 0.5 or greater, the behavior is like it used to be (I think the expected/correct behavior).

I've narrowed it down to the change having happened in ImageSharp.Drawing 2.0.0. It was OK in 1.0.0.

The call that used to not fill the polygon (from DrawPolygonExtensions):

```
 public static IImageProcessingContext DrawPolygon(
     this IImageProcessingContext source,
     Color color,
     float thickness,
     params PointF[] points) =>
     source.DrawPolygon(new SolidBrush(color), thickness, points)
```
</div>

tocsoft:
### Prerequisites

### Description
If the outline thickness is going to trigger the noop behaviour of the clipper then scale the inputs and outputs to not trigger that behaviour.

I tried just removing the test in [`PolygonOffsetter.cs`](https://github.com/SixLabors/ImageSharp.Drawing/blob/2a9ee5150addbf280801ed25392a8ec29eb8e263/src/ImageSharp.Drawing/Shapes/PolygonClipper/PolygonOffsetter.cs#L121) but for some reason  that causes very small outlines to take significantly longer to generate than scaling up and down. (not investigated why).

This causes everything to generally render now, what it renders at particularly small sizes might not be the best (not investigated) but at least we generally render something better than the solid shape.

resolves #323

### DIFF
--- a/src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs
+++ b/src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs
@@ -16,6 +16,31 @@ public static class OutlinePathExtensions
     private const EndCapStyle DefaultEndCapStyle = EndCapStyle.Butt;
 
     /// <summary>
+    /// Calculates the scaling matrixes tha tmust be applied to the inout and output paths of for successful clipping.
+    /// </summary>
+    /// <param name="width">the requested width</param>
+    /// <param name="scaleUpMartrix">The matrix to apply to the input path</param>
+    /// <param name="scaleDownMartrix">The matrix to apply to the output path</param>
+    /// <returns>The final width to use internally to outlining</returns>
+    private static float CalculateScalingMatrix(float width, out Matrix3x2 scaleUpMartrix, out Matrix3x2 scaleDownMartrix)
+    {
+        // when the thickness is below a 0.5 threshold we need to scale
+        // the source path (up) and result path (down) by a factor to ensure
+        // the offest is greater than 0.5 to ensure offsetting isn't skipped.
+        scaleUpMartrix = Matrix3x2.Identity;
+        scaleDownMartrix = Matrix3x2.Identity;
+        if (width < 0.5)
+        {
+            float scale = 1 / width;
+            scaleUpMartrix = Matrix3x2.CreateScale(scale);
+            scaleDownMartrix = Matrix3x2.CreateScale(width);
+            width = 1;
+        }
+
+        return width;
+    }
+
+    /// <summary>
     /// Generates an outline of the path.
     /// </summary>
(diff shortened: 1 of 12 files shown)


src/ImageSharp.Drawing/Shapes/OutlinePathExtensions.cs
```diff
@@ -106,7 +135,9 @@ public static IPath GenerateOutline(this IPath path, float width, ReadOnlySpan<f
             return path.GenerateOutline(width, jointStyle, endCapStyle);
         }
 
-        IEnumerable<ISimplePath> paths = path.Flatten();
+        width = CalculateScalingMatrix(width, out Matrix3x2 scaleUpMartrix, out Matrix3x2 scaleDownMartrix);
+
+        IEnumerable<ISimplePath> paths = path.Transform(scaleUpMartrix).Flatten();
```
JimBobSquarePants, line 140: Shouldn't we only do this if the path is < .5f? It seems like we're adding overhead to all outline generation now.

tocsoft: `Transform(...)` is a no op (returning the original IPath) for `Matrix3x2.Identity` which will be return for sizes >= 0.5.

JimBobSquarePants: Ah... Missed that. Looks good. 

I'm hoping that once I get the new path stoker wired up this issue will go away.

JimBobSquarePants (approved):
Lovely. Thanks!!

PR #333, opened 2024-06-01
Fix casting of  list of textruns

stefannikolei:
Directly casting is not possible. That's why using the Cast method is the easiest way

### DIFF
--- a/src/ImageSharp.Drawing/Processing/RichTextOptions.cs
+++ b/src/ImageSharp.Drawing/Processing/RichTextOptions.cs
@@ -33,7 +33,7 @@ public class RichTextOptions : TextOptions
     /// </summary>
     public new IReadOnlyList<RichTextRun> TextRuns
     {
-        get => (IReadOnlyList<RichTextRun>)base.TextRuns;
+        get => base.TextRuns.Cast<RichTextRun>().ToList();
         set => base.TextRuns = value;
     }
 


JimBobSquarePants:
Would direct casting not work if we explicitly set an empty array as the default value?

src/ImageSharp.Drawing/Processing/RichTextOptions.cs
```diff
@@ -33,7 +33,7 @@ public RichTextOptions(RichTextOptions options)
     /// </summary>
     public new IReadOnlyList<RichTextRun> TextRuns
     {
-        get => (IReadOnlyList<RichTextRun>)base.TextRuns;
+        get => base.TextRuns.Cast<RichTextRun>().ToList();
```
antonfirsov, line 36: This creates a copy and in my understanding `Cast` should also fail since the default instance is has the property assigned to `TextRun[]`  not `RichTextRun[]`.

> Would direct casting not work if we explicitly set an empty array as the default value?

That would be a better solution IMO. However it would be still easy to bring the object into an invalid state by downcasting `RichTextOptions` to `TextOptions`, which might be an unusual, but still valid use case:

```C#
RichTextOptions a = new(font);

TextOptions b = a;
a.TextRuns = new TextRun[0];
IReadOnlyList<RichTextRun> runs = a.TextRuns; // exception
```

JimBobSquarePants: I think we'd probably fix that with some sort of generic pattern in the future but for now our options are limited.

JimBobSquarePants:
Thanks @stefannikolei for this. I'm going to close it though as I've added a fix to #335 that assigns the empty array.

PR #351, opened 2025-01-09
ImageSharp v4 compatibility

JimBobSquarePants:
### Prerequisites

### Description

ImageSharp v4 is starting to become more stable now so it's a good time to migrate and switch focus to ImageSharp.Drawing v3


tocsoft:
Should we think about maybe merging the ImageSharp and ImageSharp.Drawing Repos? as part of the v4 release cycle? 

We originally split the repos because we where not ready to release a full release of Drawing in step with ImageSharp but as they are both now stable we could just merge the repos together... it'll make producing release much simpler as we won't need to do the dance releasing one then the other we have now.

It would also have some side benefits around issue and discussion management for the community as there would be a single repo for most things (Font specific issue withstanding).

If we do decide to merge repos then the follow up question might be should it be a single project/package or still live as 2 projects/packages but released in tandem?

JimBobSquarePants:
@tocsoft I'm going to go ahead and open a PR to merge the two libraries. I'll keep this open until then but have created a discussion announcing the change.

https://github.com/SixLabors/ImageSharp.Drawing/discussions/352

JimBobSquarePants:
I'm going to merge this as-is at the moment so not to blog ongoing work.

PR #384, opened 2026-03-16
[1.x] Use macos-26-intel on CI

antonfirsov:
### Prerequisites

### Description
Same as #383 but for 1.0, which I have no idea if it still builds or not.


antonfirsov:
We are not building/packaging this. Guess it's fine since drawing code is not prone to security holes.

